PR #3, opened 2021-01-31
Start a specification of the relayer cli tool

ethanfrey:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Just documentation of desired behavior.
Intended to collect all ideas on ops usage of the relayer and provide a clear spec on how to build it.


alpe:
👍  A bit late, sorry.
> Once there is an established connection, we can create multiple ports here. This takes a lit ...

should be "multiple channels" instead of "ports".

PR #13, opened 2021-02-09
Protobuf defs

### LINKED ISSUE #5: Compile IBC protobuf types / services
- **Summary**

Both queries and messages.

ethanfrey:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Closes #5

- **What is the new behavior (if this is a feature change)?**

Add scripts to download cosmos-sdk v0.41.0 protobuf definitions and compile ibc related ones into `src/codec`.
Add `codec` to git and ensure it compiles.

- **Other information**:


scripts/proto/define-proto.sh
```diff
@@ -0,0 +1,63 @@
...
+  "$THIRD_PARTY_PROTO_DIR/tendermint/crypto/keys.proto" \
+  "$THIRD_PARTY_PROTO_DIR/tendermint/crypto/proof.proto" \
+  "$THIRD_PARTY_PROTO_DIR/tendermint/libs/bits/types.proto" \
+  "$THIRD_PARTY_PROTO_DIR/tendermint/types/params.proto" \
+  "$THIRD_PARTY_PROTO_DIR/tendermint/types/types.proto" \
+  "$THIRD_PARTY_PROTO_DIR/tendermint/types/validator.proto" \
+  "$THIRD_PARTY_PROTO_DIR/tendermint/version/types.proto"
+
+# Remove unnecessary codec files
+rm -rf \
```
willclarktech, line 56: This section isn't useful right now but may need to be configured for this repo. Sometimes ts-proto is a bit eager to generate codec files that aren't needed.

.prettierignore
```diff
@@ -1,2 +1,3 @@
 # package.json is formatted by package managers, so we ignore it here
-package.json
\ No newline at end of file
+package.json
+src/codec
```
willclarktech, line 3: We actually format the codec files in CosmJS so they look consistent and are nicer to read. (We don't lint them obviously.)

ethanfrey: Ah, that is nice. I will try that.

Issue #24, opened 2021-02-11
Extend and Reorganize queries/ibc.ts


ethanfrey:
The file was copied over from @cosmjs as a starting point, but the organization doesn't match what we need. It only has about half the exposed queries as well.

I propose:

1. Use `queryRawProof` in the verified queries and make a separate subsection for them (eg. `ibc.proofs`). `clientStateWithProof` should also go in there (which is more of what we will want)
2. Remove the `unverified` namespace (that is default), and reorganize them. Some ideas: `ibc.clientState()`, `client.clientState()`, `ibc.client.clientState()`. Or if you really want to rename them to remove stutter: `ibc.client.state()` - these all would call `clientQueryService.ClientState()`
3. Add all the other grpc queries listed in the various `codec/ibc/core/*/queries.ts` files as direct grpc queries following whatever format you choose in 2

PR #25, opened 2021-02-11
Reorganise IBC extension

willclarktech:
Based off #20

Part 1 of #24

### DIFF
--- a/src/lib/queries/ibc.ts
+++ b/src/lib/queries/ibc.ts
@@ -1,5 +1,4 @@
 import { toAscii } from '@cosmjs/encoding';
-import { Uint64 } from '@cosmjs/math';
 import { createPagination, createRpc, QueryClient } from '@cosmjs/stargate';
 import Long from 'long';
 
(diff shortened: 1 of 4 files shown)


src/lib/queries/ibc.ts
```diff
@@ -356,6 +290,104 @@ export function setupIbcExtension(base: QueryClient): IbcExtension {
...
+            channelId: string,
+            sequence: number
+          ) => {
+            // key: https://github.com/cosmos/cosmos-sdk/blob/ef0a7344af345882729598bc2958a21143930a6b/x/ibc/24-host/keys.go#L183-L185
+            const key = toAscii(
+              `commitments/ports/${portId}/channels/${channelId}/packets/${sequence}`
+            );
+            const proven = await base.queryRawProof('ibc', key);
+            const commitment = proven.value;
+            const proof = ProofOps.encode(proven.proof).finish();
```
ethanfrey, line 323: Thanks for porting all these over.

Odd that there is the proof field in the grpc (which would seem to fit), but is never filled. So we make raw queries and then construct the grpc result type.

Anyway, great to have these, just grumbling about the sdk

ethanfrey (approved):
looks good

PR #31, opened 2021-02-16
Extend/tidy IBC queries

### LINKED ISSUE #24: Extend and Reorganize queries/ibc.ts

willclarktech:
Closes #24

### DIFF
--- a/src/lib/queries/ibc.ts
+++ b/src/lib/queries/ibc.ts
@@ -37,6 +37,7 @@ import {
   QueryClientImpl as ConnectionQuery,
   QueryClientConnectionsResponse,
   QueryConnectionClientStateResponse,
+  QueryConnectionConsensusStateRequest,
   QueryConnectionConsensusStateResponse,
   QueryConnectionResponse,
   QueryConnectionsResponse,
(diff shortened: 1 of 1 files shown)


src/lib/queries/ibc.ts
```diff
@@ -79,21 +102,28 @@ export interface IbcExtension {
...
       readonly consensusState: (
         clientId: string,
         height?: number
       ) => Promise<QueryConsensusStateResponse>;
+      readonly consensusStates: (
+        clientId: string,
+        paginationKey?: Uint8Array
+      ) => Promise<QueryConsensusStatesResponse>;
+      readonly params: () => Promise<QueryClientParamsResponse>;
+      readonly stateTm: (clientId: string) => Promise<TendermintClientState>;
```
ethanfrey, line 126: Ah, there it is, just reordered. It would be nice to have statesTm() in plural,but please feel free to add that with the auto-pagination

ethanfrey (approved):
ack

Issue #23, opened 2021-02-11
Make configurable fees for the ibc transactions


ethanfrey:
In `ibcclient.ts` I created a prebuilt fee table with hardcoded values 

```ts
// TODO: replace this with buildFees in IbcClient constructor to take custom gasPrice, override gas needed
const fees = {
  initClient: {
    amount: coins(2500, 'ucosm'),
    gas: '100000',
  },
  // ...
};
```

We should:

1. Allow these to be configurable in IbcClient, like the other options on SigningStargateClient. They can be a separate member of Options for just the ibc-related default gas if that is easier way to extend the options.
2. Provide sensible default gas for each transaction - take the current values and see if you can lower them a bit
3. Store the values in IbcClient and use them (remove the global fees object)

willclarktech:
Closed in #32

PR #32, opened 2021-02-16
Make IBC fees configurable

willclarktech:
Part 1 of #23

### DIFF
--- a/src/lib/ibcclient.ts
+++ b/src/lib/ibcclient.ts
@@ -1,5 +1,13 @@
 import { toAscii } from '@cosmjs/encoding';
-import { Coin, coins, logs } from '@cosmjs/launchpad';
+import {
+  buildFeeTable,
+  Coin,
+  FeeTable,
+  GasLimits,
+  GasPrice,
+  logs,
+  StdFee,
+} from '@cosmjs/launchpad';
 import { OfflineSigner, Registry } from '@cosmjs/proto-signing';
 import {
   AuthExtension,
@@ -15,7 +23,6 @@ import {
   SigningStargateClient,
   SigningStargateClientOptions,
 } from '@cosmjs/stargate';
-// import { firstEvent } from '@cosmjs/stream';
 import {
   adaptor34,
   CommitResponse,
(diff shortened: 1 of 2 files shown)


src/lib/ibcclient.ts
```diff
@@ -171,35 +179,32 @@ function createBroadcastTxErrorMessage(result: BroadcastTxFailure): string {
...
+
+const defaultGasPrice = GasPrice.fromString('0.025ucosm');
+const defaultGasLimits: GasLimits<IbcFeeTable & CosmosFeeTable> = {
+  send: 80000,
+  initClient: 100000,
+  updateClient: 400000,
+  initConnection: 100000,
+  connectionHandshake: 200000,
+  initChannel: 100000,
+  channelHandshake: 100000,
```
ethanfrey, line 203: That should be 200000.

I did some minor tune-ing of prices to get the values on the left. I think they are all within a factor of 2 (you cannot safely half them). updateClient can likely go down a bunch in tests, but if there are 100 validators, then not.

src/lib/ibcclient.ts
```diff
@@ -239,6 +248,12 @@ export class IbcClient {
       setupIbcExtension
     );
     this.senderAddress = senderAddress;
+    const { gasPrice = defaultGasPrice, gasLimits = {} } = options;
```
ethanfrey, line 251: Nice

src/lib/ibcclient.ts
```diff
@@ -171,35 +179,32 @@ function createBroadcastTxErrorMessage(result: BroadcastTxFailure): string {
...
+  readonly channelHandshake: StdFee;
+}
+
+export type IbcClientOptions = SigningStargateClientOptions & {
+  gasLimits?: Partial<GasLimits<IbcFeeTable>>;
+};
+
+const defaultGasPrice = GasPrice.fromString('0.025ucosm');
+const defaultGasLimits: GasLimits<IbcFeeTable & CosmosFeeTable> = {
+  send: 80000,
```
ethanfrey, line 197: Isn't `send` handled inside the SigningCosmosClient with it's defaults?

I bet you could reduce this to 10000 and it would not have any effect. If so, then remove it as misleading. (You can override it, but that default is not used.

ethanfrey (approved):
A few minor comments, and get the CI to pass, then good to merge

src/lib/ibcclient.ts
willclarktech: Ah yes, I should make this a `Partial<GasLimits<IbcFeeTable & CosmosFeeTable>>`

PR #37, opened 2021-02-17
Add auto-paginating methods

### LINKED ISSUE #28: Add auto-paginating variants of ibc queries
Part of #24 

There are some queries which hit their limit quickly, and if running many tests on a chain will fail to list all items, leading to test failures. Worse, they will cause the relayer to misbehave on highly loaded chains.

To solve this, let's add some query variants that get all the data using pagination. And ensure any grpc query that does accept `PageRequest` / `paginationKey` exposes that in the query interface. In particular, at least the following variants should be added:

- [ ] `ibc.channel.channels` => `allChannels`
- [ ] `ibc.channel.connectionChannels` => `allConnectionChannels`
- [ ] `ibc.client.states` => `allStates`, also `allStatesTm` which maps the above result with the decoding as in `stateTm`
- [ ] `ibc.connection.connections` => `allConnections`
- [ ] `ibc.connection.clientConnections` => `allClientConnections`
- [ ] `ibc.channel.packetCommitments` => `allPacketCommitments`
- [ ] `ibc.channel.packetAcknowledgements` => `allPacketAcknowledgements`
- [ ] `ibc.channel.unreceivedPackets` => `allUnreceivedPackets`
- [ ] `ibc.channel.unreceivedAcks` => `allUnreceivedAcks`

willclarktech:
Closes #28

### DIFF
--- a/src/lib/queries/ibc.ts
+++ b/src/lib/queries/ibc.ts
@@ -45,6 +45,15 @@ import {
 import { ClientState as TendermintClientState } from '../../codec/ibc/lightclients/tendermint/v1/tendermint';
 import { ProofOps } from '../../codec/tendermint/crypto/proof';
 
+function decodeTendermintClientStateAny(
+  clientState: Any | undefined
+): TendermintClientState {
+  if (clientState?.typeUrl !== '/ibc.lightclients.tendermint.v1.ClientState') {
+    throw new Error(`Unexpected client state type: ${clientState?.typeUrl}`);
+  }
+  return TendermintClientState.decode(clientState.value);
+}
+
 export interface IbcExtension {
   readonly ibc: {
     readonly channel: {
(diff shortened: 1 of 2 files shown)


src/lib/queries/ibc.ts
```diff
@@ -45,6 +45,15 @@ import {
 import { ClientState as TendermintClientState } from '../../codec/ibc/lightclients/tendermint/v1/tendermint';
 import { ProofOps } from '../../codec/tendermint/crypto/proof';
 
+function decodeTendermintClientStateAny(
```
ethanfrey, line 48: :+1:

src/lib/queries/ibc.ts
```diff
@@ -207,6 +237,22 @@ export function setupIbcExtension(base: QueryClient): IbcExtension {
...
+          const channels = [];
+          let response: QueryChannelsResponse;
+          let key: Uint8Array | undefined;
+          do {
+            response = await channelQueryService.Channels({
+              pagination: createPagination(key),
+            });
+            channels.push(...response.channels);
+            key = response.pagination?.nextKey;
+          } while (key);
```
ethanfrey, line 250: Hehe, it's been years since I saw a good use of `do...while`. This is one.

ethanfrey (commented):
Code looks good. I am going to use one in tests and push a commit

ethanfrey (approved):
I ran tests 5 times until ibcclient.ts: `create simapp client on wasmd` failed as there was more than one pages of clients (pre and post returned same length).

I replaced `client.states()` with `client.allStates()` and it passed. So at least manual confirmation that it works. (Adding this change as a commit to the PR)

src/lib/queries/ibc.ts
willclarktech: I needed one the other day in some Python code so it was nice to be able to actually use it here!

PR #38, opened 2021-02-17
Fix async assertions

willclarktech:
If you don’t await these you get race conditions


src/lib/link.spec.ts
```diff
@@ -34,7 +34,7 @@ test.serial('initialized connection and start channel handshake', async (t) => {
   const link = await Link.createWithNewConnections(src, dest);
 
   // reject channels with invalid ports
-  t.throwsAsync(() =>
+  await t.throwsAsync(() =>
```
ethanfrey, line 37: Ah, very good eye

ethanfrey (approved):
Nice catch

PR #39, opened 2021-02-17
Implement Link.createWithExistingConnections

### LINKED ISSUE #18: Implement find existing client/connection in Endpoint/IbcClient
This only needs the supposed connection id on both sides.

It should check to see if they really do point to the other side. Most of the logic is here: https://github.com/confio/ts-relayer/blob/main/spec/ibc-relayer.md#connect

Once that works, you could also check the header for the chains and make sure they match what is stored in the client on the remote chain

* Search for $connection
* Ensure it is open
* Get the client for that connection
    * Ensure it is tendermint client and active
    * Ensure the remote chain_id in client matches the other chain's chain_id
    * (optional) check the header in latest consensus state, ensure it matches remote chain
    * Return the last update time/height

willclarktech:
Part 1 of #18

Still to do:

- add check in `Link.createWithExistingConnections` that header matches consensus state
- test for `Link.createWithExistingConnections` with partially open channel on existing connection

### DIFF
--- a/src/lib/link.spec.ts
+++ b/src/lib/link.spec.ts
@@ -61,6 +61,19 @@ test.serial('errors when reusing an invalid connection', async (t) => {
   );
 });
 
+test.serial(`errors when reusing connections which don’t match`, async (t) => {
+  const [src, dest] = await setup();
+
+  const oldLink1 = await Link.createWithNewConnections(src, dest);
+  const connA = oldLink1.endA.connectionID;
+  const oldLink2 = await Link.createWithNewConnections(src, dest);
+  const connB = oldLink2.endB.connectionID;
+
+  await t.throwsAsync(() =>
+    Link.createWithExistingConnections(src, dest, connA, connB)
+  );
+});
+
 // constants for this transport protocol
 const ics20 = {
   // we set a new port in genesis for simapp
--- a/src/lib/link.spec.ts
+++ b/src/lib/link.spec.ts
@@ -5,6 +5,17 @@ import { Order, State } from '../codec/ibc/core/channel/v1/channel';
 import { Link } from './link';
 import { setup } from './testutils.spec';
 
+// constants for this transport protocol
+const ics20 = {
+  // we set a new port in genesis for simapp
+  srcPortId: 'custom',
+  destPortId: 'transfer',
+  version: 'ics20-1',
+  ordering: Order.ORDER_UNORDERED,
+};
+
+// createWithNewConnections
+
 test.serial('establish new client-connection', async (t) => {
   const [src, dest] = await setup();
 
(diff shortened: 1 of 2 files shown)


src/lib/link.spec.ts
```diff
@@ -20,6 +20,47 @@ test.serial('establish new client-connection', async (t) => {
...
+  const connB = oldLink.endB.connectionID;
+
+  const newLink = await Link.createWithExistingConnections(
+    src,
+    dest,
+    connA,
+    connB
+  );
+
+  // ensure the data makes sense (TODO: more?)
```
ethanfrey, line 37: I would propose creating a channel on oldLink first. When you reuse connection, ensure that channel is there.

src/lib/link.spec.ts
```diff
@@ -20,6 +20,47 @@ test.serial('establish new client-connection', async (t) => {
...
+
+  // try to update both clients, ensuring this connection is stable
+  await newLink.updateClient('A');
+  // TODO: ensure it is updated
+  await newLink.updateClient('B');
+  // TODO: ensure it is updated
+});
+
+test.serial('errors when reusing an invalid connection', async (t) => {
+  const [src, dest] = await setup();
```
ethanfrey, line 55: First, please make a connection, as above. And show bad args ate rejected even if a connection does exist.

src/lib/link.spec.ts
```diff
@@ -20,6 +20,47 @@ test.serial('establish new client-connection', async (t) => {
...
+
+  const connA = 'whatever';
+  const connB = 'unreal';
```
ethanfrey, line 58: This is a good example. I would propose 2 more.

1. Try to connect nodeA with nodeA using proper connectionid. Each side will pass local checks, but fail if we compare the client with the remote chain

ethanfrey (commented):
First round of review

src/lib/link.spec.ts
```diff
@@ -20,6 +20,60 @@ test.serial('establish new client-connection', async (t) => {
...
+    newLink.endA.clientID.startsWith('07-tendermint-'),
+    newLink.endA.clientID
+  );
+  t.assert(
+    newLink.endB.clientID.startsWith('07-tendermint-'),
+    newLink.endB.clientID
+  );
+
+  // try to update both clients, ensuring this connection is stable
+  await newLink.updateClient('A');
```
ethanfrey, line 48: You could just create a second channel in the reused connection. It should have a new id and will do a thorough check it is restored properly.

src/lib/link.ts
```diff
@@ -44,10 +44,50 @@ export class Link {
...
+    if (connectionB.clientId !== connectionA.counterparty.clientId) {
+      throw new Error(
+        `Client ID ${connectionB.clientId} for connection with ID ${connB} does not match counterparty client ID ${connectionA.counterparty.clientId} for connection with ID ${connA}`
+      );
+    }
+    const [clientStateA, clientStateB] = await Promise.all([
+      nodeA.query.ibc.client.stateTm(connectionA.clientId),
+      nodeB.query.ibc.client.stateTm(connectionB.clientId),
+    ]);
+    // TODO: What should be done with these?
```
ethanfrey, line 81: Minimally, check the remote chain id. Secondly, query the latest consensus state, and you can look for a matching header on the remote node to be thorough

ethanfrey (commented):
A few more points

src/lib/link.spec.ts
willclarktech: What's the second one?

ethanfrey: Ah, second one would be a partially open channel. Manually do init/try but don't finish the handshake (copy code from the whole flow in link).

Then give valid connection ids and valid clients, but it is in a non-open state (init or tryinit). You should ensure the connection is usable at startup

src/lib/link.ts
```diff
@@ -44,10 +44,55 @@ export class Link {
...
-    throw new Error('not yet implemented');
+    const [clientIdA, clientIdB] = await createClients(nodeA, nodeB);
+    const [
+      { connection: connectionA },
+      { connection: connectionB },
+    ] = await Promise.all([
+      nodeA.query.ibc.connection.connection(connA),
+      nodeB.query.ibc.connection.connection(connB),
+    ]);
+    if (!connectionA) {
```
ethanfrey, line 55: Nice set of checks, and use of Promise.all

src/lib/link.ts
```diff
@@ -44,10 +44,55 @@ export class Link {
...
+    if (chainIdA !== clientStateB.chainId) {
+      throw new Error(
+        `Chain ID ${chainIdA} for connection with ID ${connA} does not match remote chain ID ${clientStateA.chainId}`
+      );
+    }
+    if (chainIdB !== clientStateA.chainId) {
+      throw new Error(
+        `Chain ID ${chainIdB} for connection with ID ${connB} does not match remote chain ID ${clientStateB.chainId}`
+      );
+    }
```
ethanfrey, line 92: Can you add a //TODO here to check the header. Not needed in this PR, but it would be good to have as a triple check (assuming there is some malicious mis-configuration going on)

src/lib/link.ts
```diff
@@ -144,6 +189,7 @@ export class Link {
       src.connectionID,
       version
     );
+    console.log('HERE 1');
```
ethanfrey, line 192: To remove before merging. I can add them back when debugging the channel handshake.

src/lib/link.spec.ts
```diff
@@ -110,3 +112,89 @@ test.serial(
...
+
+  const oldLink = await Link.createWithNewConnections(src, dest);
+  const connA = oldLink.endA.connectionID;
+
+  await t.throwsAsync(() =>
+    Link.createWithExistingConnections(src, src, connA, connA)
+  );
+});
+
+test.serial(`errors when reusing connections which don’t match`, async (t) => {
```
ethanfrey, line 189: Nice one

ethanfrey (approved):
Looks good. Just comment out that broken test and leave me an issue to investigate.

Thanks

ethanfrey:
Okay debugging.

Creating the second channel with `oldLink` works fine.
Looking at createWithExistingConnections, I see the first line:

```ts
  public static async createWithExistingConnections(
    nodeA: IbcClient,
    nodeB: IbcClient,
    connA: string,
    connB: string
  ): Promise<Link> {
    const [clientIdA, clientIdB] = await createClients(nodeA, nodeB);
```

That is the error. I will adjust it

PR #67, opened 2021-02-21
Update ts-proto to 1.67 and regenerate codecs

### LINKED ISSUE #64: Use canonical ts-proto to create encodings.
After https://github.com/stephenh/ts-proto/pull/211 is merged.

Then I can remove my local hacks and we have an easy flow to regenerate the protobuf bindings from source from anyone's machine.

ethanfrey:
Closes #64 

This has the timestamp and default fields fixes in.
Strange encoding failure.

```
transfer message and send packets

  Rejected promise returned by test. Reason:

  TypeError {
    message: 'message.revisionNumber.isZero is not a function',
  }

  › Object.encode (build/main/codec/ibc/core/client/v1/client.js:2:19514)
  › Object.encode (build/main/codec/ibc/core/channel/v1/channel.js:14:24172)
  › Object.encode (build/main/codec/ibc/core/channel/v1/tx.js:2:55363)
  › Registry.encode (node_modules/@cosmjs/proto-signing/src/registry.ts:77:33)
  › node_modules/@cosmjs/proto-signing/src/registry.ts:82:33
  › Array.map (<anonymous>)
  › Registry.encodeTxBody (node_modules/@cosmjs/proto-signing/src/registry.ts:81:51)
  › Registry.encode (node_modules/@cosmjs/proto-signing/src/registry.ts:73:19)
  › SigningStargateClient.signAndBroadcast (node_modules/@cosmjs/stargate/src/signingstargateclient.ts:155:39)
```

In this case, `message.revisionNumber` is `"0"` not `Long(0)`. (From `console.log` calls, no idea how that happened)

@webmaster128 Not sure if there are any insights from https://github.com/cosmos/cosmjs/pull/683

### DIFF
--- a/src/codec/ibc/core/client/v1/genesis.ts
+++ b/src/codec/ibc/core/client/v1/genesis.ts
@@ -60,11 +60,15 @@ export const GenesisState = {
     for (const v of message.clientsMetadata) {
       IdentifiedGenesisMetadata.encode(v!, writer.uint32(26).fork()).ldelim();
     }
-    if (message.params !== undefined && message.params !== undefined) {
+    if (message.params !== undefined) {
       Params.encode(message.params, writer.uint32(34).fork()).ldelim();
     }
-    writer.uint32(40).bool(message.createLocalhost);
-    writer.uint32(48).uint64(message.nextClientSequence);
+    if (message.createLocalhost === true) {
+      writer.uint32(40).bool(message.createLocalhost);
+    }
+    if (!message.nextClientSequence.isZero()) {
+      writer.uint32(48).uint64(message.nextClientSequence);
+    }
     return writer;
   },
 
@@ -160,6 +164,40 @@ export const GenesisState = {
     return message;
   },
 
+  toJSON(message: GenesisState): unknown {
+    const obj: any = {};
+    if (message.clients) {
+      obj.clients = message.clients.map((e) =>
+        e ? IdentifiedClientState.toJSON(e) : undefined
+      );
+    } else {
+      obj.clients = [];
+    }
+    if (message.clientsConsensus) {
+      obj.clientsConsensus = message.clientsConsensus.map((e) =>
+        e ? ClientConsensusStates.toJSON(e) : undefined
+      );
+    } else {
+      obj.clientsConsensus = [];
+    }
+    if (message.clientsMetadata) {
+      obj.clientsMetadata = message.clientsMetadata.map((e) =>
+        e ? IdentifiedGenesisMetadata.toJSON(e) : undefined
+      );
+    } else {
+      obj.clientsMetadata = [];
+    }
+    message.params !== undefined &&
+      (obj.params = message.params ? Params.toJSON(message.params) : undefined);
+    message.createLocalhost !== undefined &&
+      (obj.createLocalhost = message.createLocalhost);
+    message.nextClientSequence !== undefined &&
+      (obj.nextClientSequence = (
+        message.nextClientSequence || Long.UZERO
+      ).toString());
+    return obj;
+  },
+
   fromPartial(object: DeepPartial<GenesisState>): GenesisState {
     const message = { ...baseGenesisState } as GenesisState;
     message.clients = [];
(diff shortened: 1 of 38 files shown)


webmaster128:
> In this case, `message.revisionNumber` is `"0"` not `Long(0)`. (From `console.log` calls, no idea how that happened)

Then this is the bug most likely. You set `--ts_proto_opt="esModuleInterop=true,forceLong=long,useOptionals=true,useDate=false" \`, have

```ts
const baseHeight: object = {
  revisionNumber: Long.UZERO,
  revisionHeight: Long.UZERO,
};
```

and

```ts
export interface Height {
  /** the revision that the client is currently on */
  revisionNumber: Long;
  /** the height within the given revision */
  revisionHeight: Long;
}
```

so it must be `Long` at any point. Now you need to find the source of the `"0"`.

webmaster128:
It is possible that this issues existed before and is now revealed by https://github.com/stephenh/ts-proto/pull/215.

ethanfrey:
I am pretty sure the source is in cosmjs. I print out what I pass into signAndBraodcast and that has Long.

I figured you may have stumbled upon this yourself. But I guess not. This PR is not urgent and I will take care of other issues before debugging

src/lib/utils.ts
```diff
@@ -184,32 +183,34 @@ export function parsePacket({ type, attributes }: ParsedEvent): Packet {
   );
   const [timeoutRevisionNumber, timeoutRevisionHeight] =
     attributesObj.packet_timeout_height?.split('-') ?? [];
-  return {
-    sequence: Long.fromNumber(attributesObj.packet_sequence ?? 0, true),
+  return Packet.fromPartial({
+    sequence: Long.fromString(attributesObj.packet_sequence ?? '0', true),
```
ethanfrey, line 187: The bug was here

src/lib/utils.ts
```diff
@@ -184,32 +183,34 @@ export function parsePacket({ type, attributes }: ParsedEvent): Packet {
...
+    data: attributesObj.packet_data
+      ? toUtf8(attributesObj.packet_data)
+      : undefined,
     /** block height after which the packet times out */
     timeoutHeight:
       timeoutRevisionNumber && timeoutRevisionHeight
         ? Height.fromPartial({
-            revisionNumber: timeoutRevisionNumber,
-            revisionHeight: timeoutRevisionHeight,
+            revisionNumber: Long.fromString(timeoutRevisionNumber),
```
ethanfrey, line 204: And here (passing string where it should be long)

src/lib/utils.ts
```diff
@@ -174,8 +174,7 @@ export function parsePacket({ type, attributes }: ParsedEvent): Packet {
   if (type !== 'send_packet') {
     throw new Error(`Cannot parse event of type ${type}`);
   }
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
ethanfrey, line 176: And the root cause was ignoring eslint and using "any" which didn't check the proper types.

ethanfrey (commented):
Got this working. The error was a bug in parsing that returned string.

`"5".toNumber()` worked just like `new Long(5).toNumber()` in the old ts-proto code. With the new isZero check in ts-proto, this difference is important

src/codec/ibc/core/client/v1/genesis.ts
```diff
@@ -420,7 +430,7 @@ var globalThis: any = (() => {
   if (typeof self !== 'undefined') return self;
   if (typeof window !== 'undefined') return window;
   if (typeof global !== 'undefined') return global;
-  throw new Error('Unable to locate global object');
+  throw 'Unable to locate global object';
```
willclarktech, line 433: I wonder why this was changed. 😕

src/lib/utils.ts
```diff
@@ -183,32 +191,34 @@ export function parsePacket({ type, attributes }: ParsedEvent): Packet {
   );
   const [timeoutRevisionNumber, timeoutRevisionHeight] =
     attributesObj.packet_timeout_height?.split('-') ?? [];
-  return {
-    sequence: Long.fromNumber(attributesObj.packet_sequence ?? 0, true),
+  return Packet.fromPartial({
+    sequence: may(Long.fromString, attributesObj.packet_sequence),
```
willclarktech, line 195: Do you not have to specify it's unsigned with `Long.fromString`?

src/codec/ibc/core/client/v1/genesis.ts
ethanfrey: Ask upstream. Another issue on ts-proto?

src/lib/utils.ts
ethanfrey: It doesn't matter here

PR #70, opened 2021-02-22
Query past acks

### LINKED ISSUE #44: Implement query all pending acks
Depends on #43

Perform a similar query set and return equivalent metadata. 

Note we also need the original packet when submitting an acknowledgement. I am open for suggestions if this function should do it, or if we should have a second function to take the (filtered) list of acks and get the packet data for them. If the data is easily available here, add it. If it involves queries to a second chain, then don't include it in this query.

ethanfrey:
Closes #44

### DIFF
--- a/src/lib/link.ts
+++ b/src/lib/link.ts
@@ -2,7 +2,12 @@ import { arrayContentEquals } from '@cosmjs/utils';
 
 import { Order, State } from '../codec/ibc/core/channel/v1/channel';
 
-import { Endpoint, PacketWithMetadata, QueryOpts } from './endpoint';
+import {
+  AckWithMetadata,
+  Endpoint,
+  PacketWithMetadata,
+  QueryOpts,
+} from './endpoint';
 import {
   buildCreateClientArgs,
   ChannelInfo,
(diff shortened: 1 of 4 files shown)


src/lib/link.ts
```diff
@@ -345,6 +350,31 @@ export class Link {
...
+    // TODO: handle this when there are multiple channels,
+    const toCheck = allAcks.map((ack) =>
+      ack.originalPacket.sequence.toNumber()
+    );
+    const { sequences } = await dest.client.query.ibc.channel.unreceivedAcks(
+      sourcePort,
+      sourceChannel,
+      toCheck
+    );
+    const unreceived = new Map(sequences.map((seq) => [seq.toNumber(), true]));
```
willclarktech, line 372: Why not use `new Set(sequences)` and test for membership with `unreceived.has(ack.originalPacket.sequence.toNumber())`?

willclarktech (approved):

src/lib/link.ts
ethanfrey: Can do. I did not know about that one

ethanfrey: Updated

Issue #55, opened 2021-02-18
Add ibc-setup init


ethanfrey:
Requires #52 

Implement `ibc-setup init` as [defined in spec](https://github.com/confio/ts-relayer/blob/main/spec/ibc-setup.md#init). 
Ensure all commands read `app.yaml` as well for initialization.

PR #73, opened 2021-02-23
ibc-setup base

### LINKED ISSUE #51: Implement registry parsing
Builds on #50

Start work on `ibc-setup`

* Support configurable [home directory as specified](https://github.com/confio/ts-relayer/blob/main/spec/config.md#config-file-location)
* Parse yaml file into appropriate structure
* Output json form of parsed configs (as demo to test it)

bartmacbartek:
Closes #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Build the project.
 yarn build
 ```
 
-Link the binaries so you can reference them w/o specyfing the full path. Use `npm` since [yarn has issues with setting permissions](https://github.com/yarnpkg/yarn/issues/3587).
+Link the binaries so you can reference them w/o specifying the full path. Use `npm` since [yarn has issues with setting permissions](https://github.com/yarnpkg/yarn/issues/3587).
 
 ```bash
 npm link
--- a/README.md
+++ b/README.md
@@ -32,6 +32,23 @@ When you are done, you can run the following in any console:
 ./scripts/wasmd/stop.sh
 ```
 
+### Running CLIs
+
+Build the project.
+
+```bash
+yarn build
+```
+
+Link the binaries so you can reference them w/o specyfing the full path. Use `npm` since [yarn has issues with setting permissions](https://github.com/yarnpkg/yarn/issues/3587).
+
+```bash
+npm link
+ibc-setup # should be available
+
+./build/main/binary/ibc-setup/index.js # run w/o linking
+```
+
 ### Testing
 
 You must have the local blockchains running for the tests to pass.
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "@istanbuljs/nyc-config-typescript": "^1.0.1",
     "@types/sinon": "^9.0.10",
     "@types/js-yaml": "^4.0.0",
+    "@types/sinon": "^9.0.10",
     "@typescript-eslint/eslint-plugin": "^4.0.1",
     "@typescript-eslint/parser": "^4.0.1",
     "ava": "^3.12.1",
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@ava/typescript": "^1.1.1",
     "@istanbuljs/nyc-config-typescript": "^1.0.1",
     "@types/sinon": "^9.0.10",
+    "@types/js-yaml": "^4.0.0",
     "@typescript-eslint/eslint-plugin": "^4.0.1",
     "@typescript-eslint/parser": "^4.0.1",
     "ava": "^3.12.1",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "@cosmjs/tendermint-rpc": "0.24.0-alpha.25",
     "@cosmjs/utils": "0.24.0-alpha.25",
     "@types/node": "^14.14.25",
+    "axios": "^0.21.1",
     "commander": "^7.1.0",
     "js-yaml": "^4.0.0",
     "protobufjs": "~6.10.2"
--- a/package.json
+++ b/package.json
@@ -115,5 +115,8 @@
     "exclude": [
       "**/*.spec.js"
     ]
+  },
+  "bin": {
+    "ibc-setup": "build/main/binary/ibc-setup/index.js"
   }
 }
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "@cosmjs/utils": "0.24.0-alpha.25",
     "@types/node": "^14.14.25",
     "commander": "^7.1.0",
+    "js-yaml": "^4.0.0",
     "protobufjs": "~6.10.2"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "@cosmjs/tendermint-rpc": "0.24.0-alpha.25",
     "@cosmjs/utils": "0.24.0-alpha.25",
     "@types/node": "^14.14.25",
+    "commander": "^7.1.0",
     "protobufjs": "~6.10.2"
   },
   "devDependencies": {
--- a/src/binary/ibc-setup/commands/init.spec.ts
+++ b/src/binary/ibc-setup/commands/init.spec.ts
@@ -12,16 +12,47 @@ const fsMkdirSync = sinon.stub(fs, 'mkdirSync');
 const axiosGet = sinon.stub(axios, 'get');
 const fsReadFileSync = sinon.stub(fs, 'readFileSync');
 const fsWriteFileSync = sinon.stub(fs, 'writeFileSync');
-
 const fsLstatSync = sinon.stub(fs, 'lstatSync');
 
 test.beforeEach(() => {
   sinon.reset();
+
   fsLstatSync.returns({
     ...new fs.Stats(),
     isDirectory: () => true,
     isFile: () => true,
   });
+});
+
+test('read existing registry.yaml', async (t) => {
+  const options: Options = {
+    home: '/home/user',
+    src: 'AAA',
+    dest: 'BBB',
+  };
+
+  fsExistSync.onCall(0).returns(true).onCall(1).returns(true);
+
+  const registryPath = `${options.home}/registry.yaml`;
+  const registryYaml = `
+  version: 1
+  `;
+
+  axiosGet.resolves({
+    data: registryYaml,
+  });
+  fsReadFileSync.returns(registryYaml);
+  fsWriteFileSync.returns(undefined);
+
+  await run(options);
+
+  t.assert(fsExistSync.calledWithExactly(options.home));
+  t.assert(fsExistSync.calledWithExactly(registryPath));
+  t.assert(fsMkdirSync.notCalled);
+  t.assert(axiosGet.notCalled);
+  t.assert(fsReadFileSync.calledOnceWith(registryPath));
+  t.assert(fsWriteFileSync.notCalled);
+  t.assert(consoleLog.calledWith({ version: 1 }));
 });
 
 test('initialize home directory and pull registry.yaml from remote', async (t) => {
@@ -54,36 +85,5 @@ test('initialize home directory and pull registry.yaml from remote', async (t) =
   t.assert(fsReadFileSync.calledOnceWith(registryPath));
   t.assert(fsWriteFileSync.calledOnceWith(registryPath, registryYaml));
   t.assert(consoleLog.calledWithMatch(new RegExp(`at ${options.home}`)));
-  t.assert(consoleLog.calledWith({ version: 1 }));
-});
-
-test('read existing registry.yaml', async (t) => {
-  const options: Options = {
-    home: '/home/user',
-    src: 'AAA',
-    dest: 'BBB',
-  };
-
-  fsExistSync.onCall(0).returns(true).onCall(1).returns(true);
-
-  const registryPath = `${options.home}/registry.yaml`;
-  const registryYaml = `
-  version: 1
-  `;
-
-  axiosGet.resolves({
-    data: registryYaml,
-  });
-  fsReadFileSync.returns(registryYaml);
-  fsWriteFileSync.returns(undefined);
-
-  await run(options);
-
-  t.assert(fsExistSync.calledWithExactly(options.home));
-  t.assert(fsExistSync.calledWithExactly(registryPath));
-  t.assert(fsMkdirSync.notCalled);
-  t.assert(axiosGet.notCalled);
-  t.assert(fsReadFileSync.calledOnceWith(registryPath));
-  t.assert(fsWriteFileSync.notCalled);
   t.assert(consoleLog.calledWith({ version: 1 }));
 });
--- a/src/binary/ibc-setup/commands/init.spec.ts
+++ b/src/binary/ibc-setup/commands/init.spec.ts
@@ -1,0 +1,3 @@
+import test from 'ava';
+
+import { run } from './init';
--- a/src/binary/ibc-setup/commands/init.spec.ts
+++ b/src/binary/ibc-setup/commands/init.spec.ts
@@ -12,16 +12,17 @@ const fsMkdirSync = sinon.stub(fs, 'mkdirSync');
 const axiosGet = sinon.stub(axios, 'get');
 const fsReadFileSync = sinon.stub(fs, 'readFileSync');
 const fsWriteFileSync = sinon.stub(fs, 'writeFileSync');
-const fsLstatSync = sinon.stub(fs, 'lstatSync');
+sinon.replace(
+  fs,
+  'lstatSync',
+  sinon.fake.returns({
+    isDirectory: () => true,
+    isFile: () => true,
+  })
+);
 
 test.beforeEach(() => {
   sinon.reset();
-
-  fsLstatSync.returns({
-    ...new fs.Stats(),
-    isDirectory: () => true,
-    isFile: () => true,
-  });
 });
 
 test('read existing registry.yaml', async (t) => {
--- a/src/binary/ibc-setup/commands/init.spec.ts
+++ b/src/binary/ibc-setup/commands/init.spec.ts
@@ -1,3 +1,0 @@
-import test from 'ava';
-
-import { run } from './init';
--- a/src/binary/ibc-setup/commands/init.spec.ts
+++ b/src/binary/ibc-setup/commands/init.spec.ts
@@ -47,8 +47,6 @@ test('read existing registry.yaml', async (t) => {
 
   await run(options);
 
-  t.assert(fsExistSync.calledWithExactly(options.home));
-  t.assert(fsExistSync.calledWithExactly(registryPath));
   t.assert(fsMkdirSync.notCalled);
   t.assert(axiosGet.notCalled);
   t.assert(fsReadFileSync.calledOnceWith(registryPath));
@@ -79,8 +77,6 @@ test('initialize home directory and pull registry.yaml from remote', async (t) =
 
   await run(options);
 
-  t.assert(fsExistSync.calledWithExactly(options.home));
-  t.assert(fsExistSync.calledWithExactly(registryPath));
   t.assert(fsMkdirSync.calledOnceWith(options.home));
   t.assert(axiosGet.calledOnce);
   t.assert(fsReadFileSync.calledOnceWith(registryPath));
--- a/src/binary/ibc-setup/commands/init.spec.ts
+++ b/src/binary/ibc-setup/commands/init.spec.ts
@@ -1,0 +1,89 @@
+import fs from 'fs';
+
+import test from 'ava';
+import axios from 'axios';
+import sinon from 'sinon';
+
+import { Options, run } from './init';
+
+const consoleLog = sinon.stub(console, 'log');
+const fsExistSync = sinon.stub(fs, 'existsSync');
+const fsMkdirSync = sinon.stub(fs, 'mkdirSync');
+const axiosGet = sinon.stub(axios, 'get');
+const fsReadFileSync = sinon.stub(fs, 'readFileSync');
+const fsWriteFileSync = sinon.stub(fs, 'writeFileSync');
+
+const fsLstatSync = sinon.stub(fs, 'lstatSync');
+
+test.beforeEach(() => {
+  sinon.reset();
+  fsLstatSync.returns({
+    ...new fs.Stats(),
+    isDirectory: () => true,
+    isFile: () => true,
+  });
+});
+
+test('initialize home directory and pull registry.yaml from remote', async (t) => {
+  const options: Options = {
+    home: '/home/user',
+    src: 'AAA',
+    dest: 'BBB',
+  };
+
+  fsExistSync.onCall(0).returns(false).onCall(1).returns(false);
+  fsMkdirSync.returns(options.home);
+
+  const registryPath = `${options.home}/registry.yaml`;
+  const registryYaml = `
+  version: 1
+  `;
+
+  axiosGet.resolves({
+    data: registryYaml,
+  });
+  fsReadFileSync.returns(registryYaml);
+  fsWriteFileSync.returns(undefined);
+
+  await run(options);
+
+  t.assert(fsExistSync.calledWithExactly(options.home));
+  t.assert(fsExistSync.calledWithExactly(registryPath));
+  t.assert(fsMkdirSync.calledOnceWith(options.home));
+  t.assert(axiosGet.calledOnce);
+  t.assert(fsReadFileSync.calledOnceWith(registryPath));
+  t.assert(fsWriteFileSync.calledOnceWith(registryPath, registryYaml));
+  t.assert(consoleLog.calledWithMatch(new RegExp(`at ${options.home}`)));
+  t.assert(consoleLog.calledWith({ version: 1 }));
+});
+
+test('read existing registry.yaml', async (t) => {
+  const options: Options = {
+    home: '/home/user',
+    src: 'AAA',
+    dest: 'BBB',
+  };
+
+  fsExistSync.onCall(0).returns(true).onCall(1).returns(true);
+
+  const registryPath = `${options.home}/registry.yaml`;
+  const registryYaml = `
+  version: 1
+  `;
+
+  axiosGet.resolves({
+    data: registryYaml,
+  });
+  fsReadFileSync.returns(registryYaml);
+  fsWriteFileSync.returns(undefined);
+
+  await run(options);
+
+  t.assert(fsExistSync.calledWithExactly(options.home));
+  t.assert(fsExistSync.calledWithExactly(registryPath));
+  t.assert(fsMkdirSync.notCalled);
+  t.assert(axiosGet.notCalled);
+  t.assert(fsReadFileSync.calledOnceWith(registryPath));
+  t.assert(fsWriteFileSync.notCalled);
+  t.assert(consoleLog.calledWith({ version: 1 }));
+});
--- a/src/binary/ibc-setup/commands/init.spec.ts
+++ b/src/binary/ibc-setup/commands/init.spec.ts
@@ -31,19 +31,17 @@ test('read existing registry.yaml', async (t) => {
     src: 'AAA',
     dest: 'BBB',
   };
-
-  fsExistSync.onCall(0).returns(true).onCall(1).returns(true);
-
   const registryPath = `${options.home}/registry.yaml`;
   const registryYaml = `
   version: 1
   `;
 
+  fsExistSync.onCall(0).returns(true).onCall(1).returns(true);
   axiosGet.resolves({
     data: registryYaml,
   });
   fsReadFileSync.returns(registryYaml);
-  fsWriteFileSync.returns(undefined);
+  fsWriteFileSync.returns();
 
   await run(options);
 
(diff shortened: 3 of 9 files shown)


src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const REGISTRY_FILE_PATH = path.join(options.home, REGISTRY_FILE);
+  if (!fs.existsSync(REGISTRY_FILE_PATH)) {
+    // TODO: download registry.yaml from default location
+  } else if (!fs.lstatSync(REGISTRY_FILE_PATH).isFile()) {
+    throw new Error(`${REGISTRY_FILE_PATH} must be a file. It is a directory.`);
+  }
+
+  // TODO: Read registry.yaml
+}
```
bartmacbartek, line 58: This one we'll test aka integration testing we spoke about.

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export function init(_: unknown, program: Command) {
+  const programOptions: ProgramOptions = program.opts();
+
+  // TODO: Resolve options with precedence (env, command line etc.)
+  const options = {
+    src: programOptions.src ?? 'src',
+    dest: programOptions.dest ?? 'dest',
+    home: resolveHome(programOptions.home),
+  };
```
bartmacbartek, line 37: The `ibc-setup init` will create`app.yaml` if it doesn't exist. It means, that we should use `app.yaml` while resolving options for this specific command. Is that right?

ethanfrey:
Please add closes #51 in the description, so it links to the PR

package.json
```diff
@@ -113,5 +114,8 @@
     "exclude": [
       "**/*.spec.js"
     ]
+  },
+  "bin": {
```
ethanfrey, line 118: :+1:

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  const REGISTRY_FILE_PATH = path.join(options.home, REGISTRY_FILE);
+  if (!fs.existsSync(REGISTRY_FILE_PATH)) {
+    // TODO: download registry.yaml from default location
+  } else if (!fs.lstatSync(REGISTRY_FILE_PATH).isFile()) {
+    throw new Error(`${REGISTRY_FILE_PATH} must be a file. It is a directory.`);
+  }
+
+  // TODO: Read registry.yaml
```
ethanfrey, line 57: I think this should be it's own function (when implemented), it will be used by almost every command.
The setup above (create if needed, etc) is great to have just in this `run` function.

I like the separation of external `init` and internal `run` (but maybe we will need to export `run` to cover it with tests?)

src/binary/ibc-setup/commands/init.ts
ethanfrey: Sounds good

src/binary/ibc-setup/commands/init.ts
ethanfrey: Writing that function and testing it will close #51

The rest of this closes #55

src/binary/ibc-setup/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+// TODO: fill options and commands descriptions
+program
+  .description('ibc-setup program description')
+  .option(
+    '--home <path>',
+    'home option description (default: $HOME/.ibc-setup)'
+  );
+
+program
```
ethanfrey, line 17: Nice. I love how simple this all looks

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  return `${process.env.HOME}/.ibc-setup`;
+}
+
+const REGISTRY_FILE = 'registry.yaml';
+
+export function init(_: unknown, program: Command) {
+  const programOptions: ProgramOptions = program.opts();
+
+  // TODO: Resolve options with precedence (env, command line etc.)
```
ethanfrey, line 32: Yup. I think we can add some `resolveEnv` functions later. I would propose something like this:

```ts
const home = resolveHome(programOptions.home);
// this returns `Record<string, string | undefined>`, parsing the app.yaml file in home
// if file is missing, just returns {} (no error)
const appConfig = loadAppConfig(home);
// resolveEnv takes the various items to look up. Throws an error if all are undefined (second read from env)
const src = resolveEnv(programOptions.src, 'RELAYER_SRC', appConfig.src);
const dest = resolveEnv(programOptions.dest, 'RELAYER_DEST', appConfig.dest);

const options = {src, dest, home};
```

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,58 @@
...
+import { GlobalOptions } from '../types';
+
+type Options = GlobalOptions & {
+  src: string;
+  dest: string;
+};
+
+type ProgramOptions = Partial<Options>;
+
+function resolveHome(home?: string) {
```
ethanfrey, line 15: Nice.

ethanfrey (commented):
Nice start. Added some comments about future direction.
Maybe you already planned those paths, but just to make sure we are on the same page

src/binary/ibc-setup/commands/init.ts
bartmacbartek: @ethanfrey I updated the comment since it didn't make sense before.

ethanfrey: Yes, this is correct. You can use flags or env variables here.

You could just use the [same code I mention below](https://github.com/confio/ts-relayer/pull/73#discussion_r580853285), but replace with `const appConfig = {}` (eg. missing file)

ethanfrey:
This is an odd CI error. Any idea where it comes from?

> Error: Cannot find module './polyfills.js'

https://app.circleci.com/pipelines/github/confio/ts-relayer/121/workflows/6e643b08-231e-4ba6-a483-db6ae8565511/jobs/246

src/binary/ibc-setup/commands/init.ts
bartmacbartek: The `run` will be tested for sure so it must be exported.

src/binary/ibc-setup/commands/init.ts
bartmacbartek: I thought of a similar thing, gonna do :) I just didn't touch this yet.

src/binary/ibc-setup/commands/init.ts
bartmacbartek: >I think this should be it's own function (when implemented), it will be used by almost every command.

I guess it's going to be reused but for now I just had to put it somewhere. Refactors is ahead!

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const REGISTRY_FILE = 'registry.yaml';
+
+export function init(_: unknown, program: Command) {
+  const programOptions: ProgramOptions = program.opts();
+
+  const options = {
+    src: resolveOption('src', programOptions.src, process.env.RELAYER_SRC),
+    dest: resolveOption('dest', programOptions.dest, process.env.RELAYER_DEST),
+    home: resolveOption('home', programOptions.home, getDefaultHome),
```
bartmacbartek, line 33: I'm thinking of curry-ing the `resolveOption` function so we can inject logger or any other parameters in the future. Alternatively, it could accept options in an array.

```ts
const options = {
  src: resolveOption('src', logger)(programOptions.src, process.env.RELAYER_SRC)
  ...
}
```

or

```ts
const options = {
  src: resolveOption('src', 'logger', [programOptions.src, process.env.RELAYER_SRC])
  ...
}
```

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  const REGISTRY_FILE_PATH = path.join(options.home, REGISTRY_FILE);
+  if (!fs.existsSync(REGISTRY_FILE_PATH)) {
+    // TODO: download registry.yaml from default location
+  } else if (!fs.lstatSync(REGISTRY_FILE_PATH).isFile()) {
+    throw new Error(`${REGISTRY_FILE_PATH} must be a file. It is a directory.`);
+  }
+
+  const registry = yaml.load(fs.readFileSync(REGISTRY_FILE_PATH, 'utf-8'));
```
bartmacbartek, line 54: This will throw if the file was removed while running the program or it has no appropriate permissions to be read or if yarm parsing fails. However, I think it's okay to not catch this since it produces a meaningful error and we want it to throw. Maybe later if we want to log this and re-throw?

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,57 @@
...
+  // TODO: registry file validation?
```
bartmacbartek, line 55: Should we care about the registry validation or skip it for now?

It'll be sort-of validated in the succeeding functions. But maybe we should validate it and cast it to concrete type? Thoughts?

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (!fs.existsSync(options.home)) {
+    fs.mkdirSync(options.home, { recursive: true });
+    console.log(`Initialized home directory at ${options.home}`);
+  } else if (!fs.lstatSync(options.home).isDirectory()) {
+    throw new Error(`${options.home} must be a directory. It is a file.`);
+  }
+
+  const REGISTRY_FILE_PATH = path.join(options.home, REGISTRY_FILE);
+  if (!fs.existsSync(REGISTRY_FILE_PATH)) {
+    // TODO: download registry.yaml from default location
```
bartmacbartek, line 49: I guess we want to download this from some remote registry and **not** ship with the binary, right?

Maybe for now, we can pull it from GitHub?

src/binary/ibc-setup/commands/init.ts
ethanfrey: Sounds good. We can throw any non-recoverable errors.

When we integrate the logger, I would propose wrapping all the "run" functions with something that catches any error and passes it to `log.error` before `process.exit(1)`. But that is a different issue

src/binary/ibc-setup/commands/init.ts
ethanfrey: It would be nice to assert a type and validate it.
Again, let's decide if we take this route and add later

src/binary/ibc-setup/utils/resolve-option.ts
```diff
@@ -0,0 +1,14 @@
+export function resolveOption(
+  identifier: string,
+  ...args: Array<(string | undefined) | (() => string | undefined | void)>
+) {
+  for (const option of args) {
+    const value = typeof option === 'function' ? option() : option;
```
ethanfrey, line 6: ah, nice one here. I like the choice of function or value.

src/binary/ibc-setup/utils/resolve-option.ts
```diff
@@ -0,0 +1,14 @@
+export function resolveOption(
+  identifier: string,
+  ...args: Array<(string | undefined) | (() => string | undefined | void)>
```
ethanfrey, line 3: Can we remove `| void` here? The function should return something

src/binary/ibc-setup/commands/init.ts
ethanfrey: I think it is fine as it is now. I don't see the purpose of passing in a logger to such a small function with no I/O

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+type ProgramOptions = Partial<Options>;
+
+const REGISTRY_FILE = 'registry.yaml';
+
+export function init(_: unknown, program: Command) {
+  const programOptions: ProgramOptions = program.opts();
+
+  function getDefaultHome() {
```
ethanfrey, line 22: This could go top-level, right? Or even in a `utils.ts` file.

src/binary/ibc-setup/commands/init.ts
ethanfrey: Yes. Use this link: https://raw.githubusercontent.com/confio/ts-relayer/main/demo/registry.yaml

We can change that later, but it works for now.

ethanfrey (commented):
More feedback

src/binary/ibc-setup/commands/init.ts
bartmacbartek: For sure, however, there's only one usage now and I'd like to keep things as close as possible. I'll lift it up when there'll be a need. Is that fine?

src/binary/ibc-setup/commands/init.ts
bartmacbartek: Alright, in case it's a quick refactor.

src/binary/ibc-setup/commands/init.ts
bartmacbartek: >When we integrate the logger, I would propose wrapping all the "run" functions with something that catches any error and passes it to log.error before process.exit(1)

Yeah, an error boundary would be nice!

src/binary/ibc-setup/utils/resolve-option.ts
bartmacbartek: It is intended for usages like:

```diff
function getSomething() {
  if (someCondition) {
    return something
  }

--   return undefined
}
```

so you don't have to return `undefined`. But maybe it's better to be explicit. What do you think?

src/binary/ibc-setup/commands/init.ts
bartmacbartek: Alright.

I think no matter of route we would have to do it.
4b45fc2

src/binary/ibc-setup/commands/init.ts
bartmacbartek: https://github.com/confio/ts-relayer/pull/73/commits/85e0baa59a1575ce930cfe4fa1d3d2fa91d25ddc

src/binary/ibc-setup/utils/resolve-option.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import { resolveOption } from './resolve-option';
+
+const stringOption1 = 'string option 1';
+const stringOption2 = 'string option 2';
+const functionWithString = () => 'function option';
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const functionWithVoid = () => {};
+
+test('leftmost not undefined option takes precedence', (t) => {
```
willclarktech, line 11: "not undefined" = "defined"?

src/binary/ibc-setup/utils/resolve-option.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  const option1 = resolveOption(
+    'first option', // identifier
+    undefined,
+    undefined,
+    stringOption1
+  );
+  t.is(option1, 'string option 1');
```
willclarktech, line 18: Nice to use the constants in the expectations too.

src/binary/ibc-setup/utils/resolve-option.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  const option1 = () =>
+    resolveOption(
+      'first option',
+      undefined,
+      functionWithVoid,
+      undefined,
+      undefined,
+      functionWithVoid
+    );
+  t.throws(option1, { instanceOf: Error });
```
willclarktech, line 77: In CosmJS we tend to test the error message (loosely) matches something we want.

src/binary/ibc-setup/commands/init.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+const consoleLog = sinon.stub(console, 'log');
+const fsExistSync = sinon.stub(fs, 'existsSync');
+const fsMkdirSync = sinon.stub(fs, 'mkdirSync');
+const axiosGet = sinon.stub(axios, 'get');
+const fsReadFileSync = sinon.stub(fs, 'readFileSync');
+const fsWriteFileSync = sinon.stub(fs, 'writeFileSync');
+const fsLstatSync = sinon.stub(fs, 'lstatSync');
+
+test.beforeEach(() => {
+  sinon.reset();
```
willclarktech, line 18: Ah great, `sinon` itself is now a default sandbox. Very nice.

src/binary/ibc-setup/commands/init.spec.ts
```diff
@@ -0,0 +1,89 @@
+import fs from 'fs';
+
+import test from 'ava';
+import axios from 'axios';
+import sinon from 'sinon';
+
+import { Options, run } from './init';
+
+const consoleLog = sinon.stub(console, 'log');
```
willclarktech, line 9: This will be replaced with a Winston logger, right?

src/binary/ibc-setup/commands/init.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  const registryPath = `${options.home}/registry.yaml`;
+  const registryYaml = `
+  version: 1
+  `;
+
+  axiosGet.resolves({
+    data: registryYaml,
+  });
+  fsReadFileSync.returns(registryYaml);
+  fsWriteFileSync.returns(undefined);
```
willclarktech, line 45: Can we group all the test fixtures together and all the stubbing together?

src/binary/ibc-setup/commands/init.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  await run(options);
+
+  t.assert(fsExistSync.calledWithExactly(options.home));
+  t.assert(fsExistSync.calledWithExactly(registryPath));
```
willclarktech, line 50: I'm not sure it's necessary to check these calls. We want to make sure that eg network requests aren't being made or directories aren't being created, but is it important whether or not `fs.existsSync` is called?

src/binary/ibc-setup/commands/init.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  t.assert(fsMkdirSync.calledOnceWith(options.home));
+  t.assert(axiosGet.calledOnce);
+  t.assert(fsReadFileSync.calledOnceWith(registryPath));
+  t.assert(fsWriteFileSync.calledOnceWith(registryPath, registryYaml));
+  t.assert(consoleLog.calledWithMatch(new RegExp(`at ${options.home}`)));
+  t.assert(consoleLog.calledWith({ version: 1 }));
+});
```
willclarktech, line 89: Would be nice to have tests for
- insufficient read/write permissions
- axios error
- bad yaml file

src/binary/ibc-setup/commands/init.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+const fsMkdirSync = sinon.stub(fs, 'mkdirSync');
+const axiosGet = sinon.stub(axios, 'get');
+const fsReadFileSync = sinon.stub(fs, 'readFileSync');
+const fsWriteFileSync = sinon.stub(fs, 'writeFileSync');
+const fsLstatSync = sinon.stub(fs, 'lstatSync');
+
+test.beforeEach(() => {
+  sinon.reset();
+
+  fsLstatSync.returns({
```
willclarktech, line 20: If you define this as a fake in the outer scope instead of a stub do you need to do anything in the `beforeEach` hook?

src/binary/ibc-setup/commands/init.spec.ts
bartmacbartek: That's correct.

src/binary/ibc-setup/commands/init.spec.ts
bartmacbartek: Yeah, it works nicely with fake: 9b5e421e7cb2bb7618bb5a6beeb27d37eed201ba 🎉

src/binary/ibc-setup/commands/init.spec.ts
bartmacbartek: I agree it's not important. I'm going to remove that so the whole test is more readable.

5198dee

src/binary/ibc-setup/commands/init.spec.ts
bartmacbartek: Sure! Those tests are not complete yet, it just a showcase :)

src/binary/ibc-setup/commands/init.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  await run(options);
+
+  t.assert(fsMkdirSync.notCalled);
+  t.assert(axiosGet.notCalled);
+  t.assert(fsReadFileSync.calledOnceWith(registryPath));
+  t.assert(fsWriteFileSync.notCalled);
+  t.assert(consoleLog.calledWith({ version: 1 }));
+});
+
+test('initialize home directory and pull registry.yaml from remote', async (t) => {
```
ethanfrey, line 57: Wow, nice work here stubbing this.
I would just made a temp dir with some files and ran it.

Of course, that approach can never actually test most error cases, only happy path. It is definitely nice to show that this is testable.

src/binary/ibc-setup/commands/init.spec.ts
ethanfrey: I think the goal here is to show how to test. I would be happy for one failure (eg. axios error) now, and let's get `init` merged. Much of the work here is to define a path, not ensure we have checked every box (yet).

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,63 @@
...
+import { GlobalOptions } from '../types';
+import { resolveOption } from '../utils/resolve-option';
+import axios from 'axios';
+
+export type Options = GlobalOptions & {
+  src: string;
+  dest: string;
+};
+
+type ProgramOptions = Partial<Options>;
```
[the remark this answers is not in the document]
ethanfrey: Agree, you can just use `Partial<Options>` there in `init` (the type is used one place)

src/binary/ibc-setup/commands/init.ts
ethanfrey: Sure. Let's get this working. There will be a few rounds of refactoring as you add some more commands

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export async function run(options: Options) {
+  if (!fs.existsSync(options.home)) {
+    fs.mkdirSync(options.home, { recursive: true });
+    console.log(`Initialized home directory at ${options.home}`);
+  } else if (!fs.lstatSync(options.home).isDirectory()) {
+    throw new Error(`${options.home} must be a directory. It is a file.`);
+  }
+
+  const REGISTRY_FILE_PATH = path.join(options.home, REGISTRY_FILE);
```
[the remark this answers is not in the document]
ethanfrey: Time for a code style guide.

I have no opinion here but let's pick something for this repo and be consistent. There is some inertia for sticking close to CosmJS, but no need to do so.

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (!fs.existsSync(REGISTRY_FILE_PATH)) {
+    const registryFromRemote = await axios.get(
+      'https://raw.githubusercontent.com/confio/ts-relayer/main/demo/registry.yaml'
```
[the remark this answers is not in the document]
ethanfrey: I suggested this. We could pin it at some point, but more likely find a better location,

src/binary/ibc-setup/utils/resolve-option.ts
ethanfrey: My feel was that it would allow `() => {}` which returns void, but totally not acceptable. But I see your point.

Let's leave it, and only tighten it up if we shoot ourselves in the foot later. That is all types are in the end, guardrails to keep us safe when coding later.

ethanfrey (commented):
Looking good. Just some minor nits.

I wouldn't spend too much time testing elements beyond showing it is possible. This framework seems a quite nice base to me. It would be good to start implementing the rest of the logic, unless there is an objection with the structure.

README.md
```diff
@@ -32,6 +32,23 @@ When you are done, you can run the following in any console:
...
 
+### Running CLIs
+
+Build the project.
+
+```bash
+yarn build
+```
+
+Link the binaries so you can reference them w/o specyfing the full path. Use `npm` since [yarn has issues with setting permissions](https://github.com/yarnpkg/yarn/issues/3587).
```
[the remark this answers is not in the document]
bartmacbartek: 92eb28b

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,63 @@
...
+  src: string;
+  dest: string;
+};
+
+type ProgramOptions = Partial<Options>;
+
+const REGISTRY_FILE = 'registry.yaml';
+
+export function init(_: unknown, program: Command) {
+  const programOptions: ProgramOptions = program.opts();
```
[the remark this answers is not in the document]
bartmacbartek: Yes, you're right. I messed sth up

I changed the interface so it now accepts flags and is not bonded with the commander: 7bacc91

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,63 @@
...
+export type Options = GlobalOptions & {
+  src: string;
+  dest: string;
+};
+
+type ProgramOptions = Partial<Options>;
+
+const REGISTRY_FILE = 'registry.yaml';
+
+export function init(_: unknown, program: Command) {
```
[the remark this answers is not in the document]
bartmacbartek: I was following the naming from "convention" from commander.js.

Anyway, it's changed now: https://github.com/confio/ts-relayer/pull/73#discussion_r582830753

src/binary/ibc-setup/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  .command('init')
+  .description('init command description')
+  .option('--src <chain>')
+  .option('--dest <chain>')
+  .action(init);
+
+program.parse();
```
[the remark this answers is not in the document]
bartmacbartek: Sure: a894442

Why do we need `.parseAsync()` now?

src/binary/ibc-setup/commands/init.spec.ts
bartmacbartek: 55bbb32

src/binary/ibc-setup/types.ts
```diff
@@ -0,0 +1,3 @@
+export type GlobalOptions = {
+  home: string;
```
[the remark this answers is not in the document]
bartmacbartek: fc70f77

src/binary/ibc-setup/index.ts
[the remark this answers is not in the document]
willclarktech: I'm not sure how `parseAsync` works, but I feel a bit uncomfortable about calling this async `run` in the synchronous command `init`. I'm hoping `parseAsync` does something nice about this based on the name?

src/binary/ibc-setup/commands/init.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  fsReadFileSync.returns('');
+  fsWriteFileSync.returns();
+
+  const error = await t.throwsAsync(async () => await run(options), {
+    instanceOf: Error,
+  });
+
+  t.assert(fsMkdirSync.calledOnceWith(options.home));
+  t.assert(axiosGet.calledOnce);
+  t.assert(error.message.includes('Cannot fetch registry.yaml from remote.'));
```
ethanfrey, line 103: Nice one

ethanfrey (approved):
This looks good as is to close #51 

If you want to make it ready for final review, we could merge and then make a separate Pr for #55 (the full init)

src/binary/ibc-setup/commands/init.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  run(options);
+}
+
+export async function run(options: Options) {
+  if (!fs.existsSync(options.home)) {
+    fs.mkdirSync(options.home, { recursive: true });
+    console.log(`Initialized home directory at ${options.home}`);
+  } else if (!fs.lstatSync(options.home).isDirectory()) {
+    throw new Error(`${options.home} must be a directory. It is a file.`);
```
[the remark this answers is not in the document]
bartmacbartek: I like it :)
c526418

src/binary/ibc-setup/utils/resolve-option.spec.ts
bartmacbartek: 6bff745
🤯

src/binary/ibc-setup/utils/resolve-option.spec.ts
bartmacbartek: agree 05f0c60

src/binary/ibc-setup/utils/resolve-option.spec.ts
bartmacbartek: Found it! ac382d9

src/binary/ibc-setup/commands/init.spec.ts
bartmacbartek: actually improved: 75a8596

src/binary/ibc-setup/index.ts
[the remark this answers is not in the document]
bartmacbartek: If we'd like to wait for it, then `parseAsync` is a way to go:


```ts
async function main() {
  const program = new Command();
  
  ...
  
  await program.parseAsync();
  
  // do something else
}

main()
```

but this is not our case.

ethanfrey (approved):
Nice start

PR #74, opened 2021-02-23
Add logger to lib

### LINKED ISSUE #57: Add logger support to library
We should accept a [winstonjs](https://github.com/winstonjs/winston) compatible logger interface in Link, Endpoint, and IbcClient and produce debug/info/warn/error logging. Verbose should be every rpc call we make (debug with the entire payload), info the general actions token ("create connection musselnet {07-tendermint-5:connection-3} <=> bifrost {07-tendermint-2:connection-1}"), and warn/error only when there are problems.

Question whether the logger should be passed in every method call, or we should set it in the constructors.

Interface should be a subset of [`winstonjs.Logger`](https://github.com/winstonjs/winston/blob/master/index.d.ts#L89-L145). We can expand over time, but harder to reduce. My proposal is something like this (taking only one aspect of `LeveledLogMethod`):

```ts
interface Logger {
  error: (message: string, ...meta: any[]) => Logger,
  warn: (message: string, ...meta: any[]) => Logger,
  info: (message: string, ...meta: any[]) => Logger,
  verbose: (message: string, ...meta: any[]) => Logger,
  debug: (message: string, ...meta: any[]) => Logger,
}
```

Maybe we want to expose the `isDebugEnabled(): boolean`... methods as well to avoid doing unneeded calculations for the logger?

willclarktech:
Part 1 of #57

Here's my first attempt:
- Pass the logger to the relevant constructor (must be handled separately for `Link` and `IbcClient` because the clients are created before the link is)
- Make it optional
- Use spies for testing (ava doesn't seem to have built-in spies so I just used sinon)
- Does `Endpoint` really need logging? It seems like all it does is pass something down to the client which will also log
- `.logger` is a public property for ease of testing, but we could make it private and open it up in the tests

### DIFF
--- a/src/lib/ibcclient.spec.ts
+++ b/src/lib/ibcclient.spec.ts
@@ -5,7 +5,14 @@ import { MsgTransfer } from '../codec/ibc/applications/transfer/v1/tx';
 
 import { buildCreateClientArgs, prepareConnectionHandshake } from './ibcclient';
 import { Link } from './link';
-import { ics20, randomAddress, setup, simapp, wasmd } from './testutils.spec';
+import {
+  ics20,
+  randomAddress,
+  setup,
+  simapp,
+  TestLogger,
+  wasmd,
+} from './testutils.spec';
 import {
   buildClientState,
   buildConsensusState,
@@ -15,12 +22,15 @@ import {
 } from './utils';
 
 test.serial('create simapp client on wasmd', async (t) => {
-  const [src, dest] = await setup();
+  const logger = new TestLogger();
+  const [src, dest] = await setup(logger);
 
   const preClients = await dest.query.ibc.client.allStates();
   const preLen = preClients.clientStates.length;
 
   const header = await src.latestHeader();
+  t.assert(logger.info.calledOnce, logger.info.callCount.toString());
+
   const conState = buildConsensusState(header);
   const cliState = buildClientState(
     await src.getChainId(),
@@ -155,9 +165,10 @@ test.serial('perform connection handshake', async (t) => {
 });
 
 test.serial('transfer message and send packets', async (t) => {
+  const logger = new TestLogger();
   // set up ics20 channel
   const [nodeA, nodeB] = await setup();
-  const link = await Link.createWithNewConnections(nodeA, nodeB);
+  const link = await Link.createWithNewConnections(nodeA, nodeB, logger);
   const channels = await link.createChannel(
     'A',
     ics20.srcPortId,
@@ -220,9 +231,10 @@ test.serial('transfer message and send packets', async (t) => {
 });
 
 test.serial('tests parsing with multi-message', async (t) => {
+  const logger = new TestLogger();
   // set up ics20 channel
   const [nodeA, nodeB] = await setup();
-  const link = await Link.createWithNewConnections(nodeA, nodeB);
+  const link = await Link.createWithNewConnections(nodeA, nodeB, logger);
   const channels = await link.createChannel(
     'A',
     ics20.srcPortId,
--- a/src/lib/ibcclient.spec.ts
+++ b/src/lib/ibcclient.spec.ts
@@ -233,7 +233,7 @@ test.serial('transfer message and send packets', async (t) => {
 test.serial('tests parsing with multi-message', async (t) => {
   const logger = new TestLogger();
   // set up ics20 channel
-  const [nodeA, nodeB] = await setup();
+  const [nodeA, nodeB] = await setup(logger);
   const link = await Link.createWithNewConnections(nodeA, nodeB, logger);
   const channels = await link.createChannel(
     'A',
@@ -251,6 +251,15 @@ test.serial('tests parsing with multi-message', async (t) => {
   const { logs: sendLogs } = await nodeA.sendTokens(srcAddr, [
     { amount: '5000', denom: simapp.denomFee },
   ]);
+  t.assert(
+    logger.verbose.calledWithMatch(/Send tokens to/),
+    logger.verbose.callCount.toString()
+  );
+  t.assert(
+    logger.debug.calledWithMatch(/Send tokens:/),
+    logger.debug.callCount.toString()
+  );
+
   const sendPackets = parsePacketsFromLogs(sendLogs);
   t.is(sendPackets.length, 0);
 
--- a/src/lib/ibcclient.spec.ts
+++ b/src/lib/ibcclient.spec.ts
@@ -29,7 +29,7 @@ test.serial('create simapp client on wasmd', async (t) => {
   const preLen = preClients.clientStates.length;
 
   const header = await src.latestHeader();
-  t.assert(logger.info.calledOnce, logger.info.callCount.toString());
+  t.assert(logger.verbose.calledOnce, logger.verbose.callCount.toString());
 
   const conState = buildConsensusState(header);
   const cliState = buildClientState(
--- a/src/lib/ibcclient.ts
+++ b/src/lib/ibcclient.ts
@@ -73,6 +73,7 @@ import {
   buildClientState,
   buildConsensusState,
   createBroadcastTxErrorMessage,
+  Logger,
   mapRpcPubKeyToProto,
   multiplyFees,
   timestampFromDateNanos,
@@ -185,6 +186,7 @@ export interface IbcFeeTable extends FeeTable {
 
 export type IbcClientOptions = SigningStargateClientOptions & {
   gasLimits?: Partial<GasLimits<IbcFeeTable>>;
+  logger?: Logger;
 };
 
 const defaultGasPrice = GasPrice.fromString('0.025ucosm');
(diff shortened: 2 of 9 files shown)


src/lib/ibcclient.ts
```diff
@@ -210,6 +212,7 @@ export class IbcClient {
     IbcExtension;
   public readonly tm: TendermintClient;
   public readonly senderAddress: string;
+  public readonly logger?: Logger;
```
bartmacbartek, line 215: Why not make it required?

willclarktech: Mostly because I would have had to update many more irrelevant lines haha. Not sure if there's a use case for no logger though?

ethanfrey: I would set it to a default `NoopLogger` implementation in the IbcClient constructor if none was provided. This means, we can always trust it to exist and call it (which will lave many more lines later), but it does nothing by default.

ethanfrey:
> Does Endpoint really need logging? It seems like all it does is pass something down to the client which will also log

No, IbcClient does debug/verbose level. Link can do info (or nothing), Endpoint is just a helper

ethanfrey:
> .logger is a public property for ease of testing, but we could make it private and open it up in the tests

I prefer everything public for now in the library, and we can tighten it down later.

src/lib/utils.ts
```diff
@@ -18,6 +18,17 @@ import {
 } from '../codec/ibc/lightclients/tendermint/v1/tendermint';
 import { PublicKey as ProtoPubKey } from '../codec/tendermint/crypto/keys';
 
+// Subset of methods from winston.js
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export interface Logger {
+  error: (message: string, ...meta: any[]) => Logger;
```
ethanfrey, line 24: I copied this from WinstonJS, but it would be good to document how these expected to be logged. (just copy the relevent lines from winstonjs)

src/lib/testutils.spec.ts
```diff
@@ -103,7 +128,8 @@ export async function queryClient(opts: QueryOpts): Promise<StargateClient> {
 
 export async function signingClient(
   opts: SigningOpts,
-  mnemonic: string
+  mnemonic: string,
+  logger?: Logger
```
ethanfrey, line 132: :+1:

src/lib/testutils.spec.ts
```diff
@@ -5,10 +5,35 @@ import { Decimal } from '@cosmjs/math';
...
 import { StargateClient } from '@cosmjs/stargate';
 import test from 'ava';
+import sinon, { SinonSpy } from 'sinon';
 
 import { Order } from '../codec/ibc/core/channel/v1/channel';
 
 import { IbcClient, IbcClientOptions } from './ibcclient';
+import { Logger } from './utils';
+
+export class TestLogger implements Logger {
```
ethanfrey, line 15: :+1:

src/lib/link.ts
```diff
@@ -217,14 +220,15 @@ export class Link {
...
+    return new Link(endA, endB, logger);
   }
 
   // you can use this if you already have the info out of bounds
   // TODO; check the validity of that data?
-  public constructor(endA: Endpoint, endB: Endpoint) {
+  public constructor(endA: Endpoint, endB: Endpoint, logger?: Logger) {
     this.endA = endA;
     this.endB = endB;
+    this.logger = logger;
```
ethanfrey, line 231: Let do `this.logger = logger ?? new NoopLogger()`

src/lib/ibcclient.ts
```diff
@@ -246,19 +249,21 @@ export class IbcClient {
...
     );
+    this.logger = logger;
   }
 
   public getChainId(): Promise<string> {
     return this.sign.getChainId();
   }
 
   public async latestHeader(): Promise<RpcHeader> {
+    this.logger?.info('Getting latest header');
```
ethanfrey, line 266: I guess we can use `?` everywhere instead of NoopLogger. I guess not much writing

src/lib/ibcclient.ts
```diff
@@ -185,6 +186,7 @@ export interface IbcFeeTable extends FeeTable {
 
 export type IbcClientOptions = SigningStargateClientOptions & {
   gasLimits?: Partial<GasLimits<IbcFeeTable>>;
+  logger?: Logger;
```
ethanfrey, line 189: Nice to add to options

src/lib/ibcclient.ts
willclarktech: Cool, I'll add one.

ethanfrey (commented):
Nice start. Some feedback

src/lib/ibcclient.ts
willclarktech: I have no preference here, it's just a matter of style.

src/lib/utils.ts
willclarktech: I'm not sure what you mean here. Are you talking about [logging levels](https://github.com/winstonjs/winston/tree/v3.3.3#logging-levels)?

ethanfrey: Sorry, I meant `...meta: any[]`.

ethanfrey: I think this is an example of using those: https://github.com/winstonjs/winston#string-interpolation

src/lib/logger.ts
```diff
@@ -0,0 +1,26 @@
...
+/* Logger interface with a subset of methods from winston.js */
+export interface Logger {
+  error: (message: string, ...meta: any[]) => Logger;
+  warn: (message: string, ...meta: any[]) => Logger;
+  info: (message: string, ...meta: any[]) => Logger;
+  verbose: (message: string, ...meta: any[]) => Logger;
+  debug: (message: string, ...meta: any[]) => Logger;
+}
+
+export class NoopLogger implements Logger {
```
ethanfrey, line 29: :+1:

src/lib/logger.ts
```diff
@@ -1,20 +1,37 @@
...
+// From https://github.com/winstonjs/winston/blob/v3.3.3/index.d.ts#L53
+type LogCallback = (
+  error?: any,
+  level?: string,
+  message?: string,
+  meta?: any
+) => void;
+
+// From https://github.com/winstonjs/winston/blob/v3.3.3/index.d.ts#L69-L75
+export interface LeveledLogMethod {
```
ethanfrey, line 12: I hesitate to add all these methods we don't know how to use.
Using this interface is fine, but I would trim it down to one or two methods and you can document how to use them

ethanfrey (commented):
Let's just use a few ways of logging and document. While trying to figure out what the meta-parameter did, I found this nice into to Winston.js: https://www.darraghoriordan.com/2019/07/30/winston-log-full-object/

In short, just pass full objects for these `...meta: any[]` args and they will get logged. Maybe this example helps: `logger.info('The suspect item!', { theitem: item })` 

It would actually be nice to try out that code sample (modifying it a bit) and find the best way to log extra data (like url, full message objects, etc). Then expose maybe one simple and one complex way to call these methods with an example.

Remember, you need to make it easy for anyone working on the library to know how to dump the proper data that will be useful later.

ethanfrey:
Besides my last nit-picks, I think this is good.

Happy for you to test out the formats in Winston a bit by hand, and then add some real log messages all over ibcclient.ts

src/lib/link.ts
```diff
@@ -256,8 +256,20 @@ export class Link {
     ordering: Order,
     version: string
   ): Promise<ChannelPair> {
+    this.logger.info(
+      `Create channel with sender ${sender}: ${srcPort} => ${destPort}`
+    );
     const { src, dest } = this.getEnds(sender);
 
+    this.logger.verbose(`Channel Open Init`);
```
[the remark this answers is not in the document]
ethanfrey: We can use source chain ID in ibcclient if you wish, but the sender ('A' or 'B') is already logged above

ethanfrey (approved):
Looks good.
Let's get this in there and make another PR to add all the log messages

src/lib/ibcclient.spec.ts
```diff
@@ -251,6 +251,15 @@ test.serial('tests parsing with multi-message', async (t) => {
   const { logs: sendLogs } = await nodeA.sendTokens(srcAddr, [
     { amount: '5000', denom: simapp.denomFee },
   ]);
+  t.assert(
+    logger.verbose.calledWithMatch(/Send tokens/),
+    logger.verbose.callCount.toString()
+  );
+  t.assert(
+    logger.debug.calledWithMatch(/Send tokens/),
```
ethanfrey, line 259: Shouldn't these be somewhat different strings to test well?

verbose: "Send tokens to"
debug: "Send tokens:"

ethanfrey (approved):
One minor point, but otherwise good to merge

ethanfrey (approved):
Let's merge this!

PR #76, opened 2021-02-24
Use logger in Link and IbcClient

### LINKED ISSUE #57: Add logger support to library

willclarktech:
Closes #57

Merge after #74

Logging level is `info` in `Link`, and `verbose`/`debug` in `IbcClient` except for creating
connections which is `info`.

### DIFF
--- a/src/lib/ibcclient.ts
+++ b/src/lib/ibcclient.ts
@@ -259,6 +259,7 @@ export class IbcClient {
   }
 
   public getChainId(): Promise<string> {
+    this.logger.verbose('Get chain ID');
     return this.sign.getChainId();
   }
 
@@ -270,6 +271,7 @@ export class IbcClient {
   }
 
   public async header(height: number): Promise<RpcHeader> {
+    this.logger.verbose(`Get header for height ${height}`);
     // TODO: expose header method on tmClient and use that
     const resp = await this.tm.blockchain(height, height);
     return resp.blockMetas[0].header;
@@ -282,6 +284,9 @@ export class IbcClient {
   }
 
   public getCommit(height?: number): Promise<CommitResponse> {
+    this.logger.verbose(
+      height === undefined ? 'Get commit' : `Get commit for height ${height}`
+    );
     return this.tm.commit(height);
   }
 
@@ -323,6 +328,7 @@ export class IbcClient {
   }
 
   public async getValidatorSet(height: number): Promise<ValidatorSet> {
+    this.logger.verbose(`Get validator set for height ${height}`);
     // we need to query the header to find out who the proposer was, and pull them out
     const { proposerAddress } = await this.header(height);
     const validators = await this.tm.validators(height);
@@ -541,6 +547,11 @@ export class IbcClient {
     msgs: EncodeObject[],
     fees: StdFee
   ): Promise<MsgResult> {
+    this.logger.verbose(`Broadcast multiple msgs`);
+    this.logger.debug(`Multiple msgs:`, {
+      msgs,
+      fees,
+    });
     const senderAddress = this.senderAddress;
     const result = await this.sign.signAndBroadcast(senderAddress, msgs, fees);
     if (isBroadcastTxFailure(result)) {
@@ -558,6 +569,7 @@ export class IbcClient {
     clientState: TendermintClientState,
     consensusState: TendermintConsensusState
   ): Promise<CreateClientResult> {
+    this.logger.verbose(`Create Tendermint client`);
     const senderAddress = this.senderAddress;
     const createMsg = {
       typeUrl: '/ibc.core.client.v1.MsgCreateClient',
@@ -573,6 +585,7 @@ export class IbcClient {
         },
       }),
     };
+    this.logger.debug('MsgCreateClient', createMsg);
 
     const result = await this.sign.signAndBroadcast(
       senderAddress,
@@ -601,6 +614,7 @@ export class IbcClient {
     clientId: string,
     header: TendermintHeader
   ): Promise<MsgResult> {
+    this.logger.verbose(`Update Tendermint client ${clientId}`);
     const senderAddress = this.senderAddress;
     const updateMsg = {
       typeUrl: '/ibc.core.client.v1.MsgUpdateClient',
@@ -613,6 +627,7 @@ export class IbcClient {
         },
       }),
     };
+    this.logger.debug(`MsgUpdateClient`, updateMsg);
 
     const result = await this.sign.signAndBroadcast(
       senderAddress,
@@ -634,6 +649,7 @@ export class IbcClient {
     clientId: string,
     remoteClientId: string
   ): Promise<CreateConnectionResult> {
+    this.logger.info(`Connection open init: ${clientId} => ${remoteClientId}`);
     const senderAddress = this.senderAddress;
     const msg = {
       typeUrl: '/ibc.core.connection.v1.MsgConnectionOpenInit',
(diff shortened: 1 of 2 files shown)


src/lib/ibcclient.ts
```diff
@@ -277,6 +284,9 @@ export class IbcClient {
   }
 
   public getCommit(height?: number): Promise<CommitResponse> {
+    this.logger.verbose(
+      height === undefined ? 'Get commit' : `Get commit for height ${height}`
```
ethanfrey, line 288: -> `undefined ? 'Get latest commit' : ...`

src/lib/ibcclient.ts
```diff
@@ -529,6 +547,11 @@ export class IbcClient {
     msgs: EncodeObject[],
     fees: StdFee
   ): Promise<MsgResult> {
+    this.logger.verbose(`Broadcast multiple msgs`);
```
ethanfrey, line 550: :+1:

src/lib/ibcclient.ts
```diff
@@ -663,6 +692,7 @@ export class IbcClient {
     myClientId: string,
     proof: ConnectionHandshakeProof
   ): Promise<CreateConnectionResult> {
+    this.logger.info(`Connection open try for my client ${myClientId}`);
```
ethanfrey, line 695: Listing at least (1) counterparty clientId and maybe (2) counterparty connection id would be nice.
Or maybe we log both connection IDs shortly before return? (We parse the newly defined one there)

src/lib/ibcclient.ts
```diff
@@ -721,6 +752,7 @@ export class IbcClient {
     myConnectionId: string,
     proof: ConnectionHandshakeProof
   ): Promise<MsgResult> {
+    this.logger.info(`Connection open ack for my connection ${myConnectionId}`);
```
ethanfrey, line 755: Yes. Just connections here, not clients (we printed the connectionIDs before.

Let's add both connection IDs: `Connection open ack for ${myConnectionId} => ${proof.connectionId}`

src/lib/ibcclient.ts
```diff
@@ -622,6 +649,7 @@ export class IbcClient {
     clientId: string,
     remoteClientId: string
   ): Promise<CreateConnectionResult> {
+    this.logger.info(`Connection open init: ${clientId} => ${remoteClientId}`);
```
ethanfrey, line 652: I like this format. Let's use a similar one (myId => remoteId) for all the connection handshake logs (and channel handshake too)

src/lib/ibcclient.ts
```diff
@@ -801,6 +836,7 @@ export class IbcClient {
     connectionId: string,
     version: string
   ): Promise<CreateChannelResult> {
+    this.logger.verbose(`Channel open init: ${portId} => ${remotePortId}`);
```
ethanfrey, line 839: Maybe log the newly created channelID with `this.logger.debug` call right before the return?

src/lib/ibcclient.ts
```diff
@@ -850,6 +887,7 @@ export class IbcClient {
     counterpartyVersion: string,
     proof: ChannelHandshake
   ): Promise<CreateChannelResult> {
+    this.logger.verbose(`Channel open try for port ${portId}`);
```
ethanfrey, line 890: Let's show both ports (and the remote channel), as in the ChannelOpenInit call

src/lib/ibcclient.ts
```diff
@@ -899,6 +938,9 @@ export class IbcClient {
     counterpartyVersion: string,
     proof: ChannelHandshake
   ): Promise<MsgResult> {
+    this.logger.verbose(
+      `Channel open ack for port ${portId}: ${channelId} => ${counterpartyChannelId}`
```
ethanfrey, line 942: :+1:

src/lib/ibcclient.ts
```diff
@@ -935,6 +978,9 @@ export class IbcClient {
     channelId: string,
     proof: ChannelHandshake
   ): Promise<MsgResult> {
+    this.logger.verbose(
+      `Chanel open confirm for port ${portId} and channel ${channelId}`
```
ethanfrey, line 982: :+1:

src/lib/ibcclient.ts
```diff
@@ -1110,6 +1163,7 @@ export class IbcClient {
     timeoutBlock?: number,
     timeoutTime?: number
   ): Promise<MsgResult> {
+    this.logger.verbose(`Transfer tokens to ${receiver}`);
```
ethanfrey, line 1166: Is there a nice stringify method for Coin?

`Transfer ${prettyToken(token)} to ${receiver}`

ethanfrey: Actually, this will only be used in tests likely, so no big deal

src/lib/link.spec.ts
```diff
@@ -4,12 +4,20 @@ import test from 'ava';
...
+  ics20,
+  randomAddress,
+  setup,
+  simapp,
+  TestLogger,
+  wasmd,
+} from './testutils.spec';
 
 test.serial('establish new client-connection', async (t) => {
+  const logger = new TestLogger();
```
ethanfrey, line 17: So, if I want detailed logs from this method, I would just use `const logger = new TestLogger(true)` ?

src/lib/link.ts
```diff
@@ -258,6 +264,9 @@ export class Link {
     source: Side,
     minHeight: number
   ): Promise<number> {
+    this.logger.info(
+      `Update client for source ${source} to min height ${minHeight}`
```
ethanfrey, line 268: This will call the above method (with log) if we need to update.
Maybe, `this.logger.info(`Checking if client for source ${source} >= height ${minHeight}` or something similar.

src/lib/link.ts
```diff
@@ -406,6 +419,7 @@ export class Link {
 
   // Returns the last height that this side knows of the other blockchain
   public async lastKnownHeader(side: Side): Promise<number> {
+    this.logger.info(`Get last known header for side ${side}`);
```
ethanfrey, line 422: This should be verbose.
Actually, I think most of the queries (besides the updateClient ones) should be on verbose level

ethanfrey (approved):
Looks good.

Can you do a little cleanup and a rebase, then good to merge.

src/lib/link.spec.ts
willclarktech: Yes, passing `true` will turn on console logging for all levels. Does it need to be configurable? I was imagining that if we wanted to see what was being logged while running tests it would be temporary and we'd want to see everything.

ethanfrey: That sounds good. And we can enable it on a per-test basis.

Seems like a good design, I just wanted to ensure I understood

PR #78, opened 2021-02-24
Add yarn.lock

### LINKED ISSUE #77: Commit yarn.lock to the repo
Why? https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

⚠️ It seems that all the future builds **with no cache** are broken because `yarn build` is failing. It's related to a different dependency tree than in the previous build. It seems like some dependency (types) released a new version with breaking change and since we don't lock the versions, it exploded:  https://app.circleci.com/pipelines/github/confio/ts-relayer/145/workflows/234297c6-fe84-4112-a03f-b52a28461013/jobs/300

To reproduce the error locally:
- remove `yarn.lock` and `node_modules` on your machine
- run
  ```bash
  yarn
  yarn build
  ```

Simply, committing `yarn.lock` would help.

bartmacbartek:
Closes #77


ethanfrey (approved):
Sure. I was trying to avoid poluting the commit history early. Makes sense to add it

PR #80, opened 2021-02-25
Support packet queries with multiple channels

### LINKED ISSUE #66: Handle multiple channels with ack/packet queries
Follow up from #65

Let's do this for acks also. Need test cases

We can create 2 ics20 channels (different IDs) on the same connection. Send some packets on each, then try the `getPendingPackets` and `relayPackets` commands to ensure this works.

willclarktech:
Closes #66

### DIFF
--- a/src/lib/link.spec.ts
+++ b/src/lib/link.spec.ts
@@ -286,7 +286,7 @@ test.serial('submit multiple tx, get unreceived packets', async (t) => {
   t.deepEqual(postAcks[0], acks[1]);
 });
 
-test.serial(
+test.serial.only(
   'submit multiple tx on multiple channels, get unreceived packets',
   async (t) => {
     const logger = new TestLogger();
@@ -307,6 +307,7 @@ test.serial(
       ics20.ordering,
       ics20.version
     );
+    t.not(channels1.src.channelId, channels2.src.channelId);
 
     // no packets here
     const noPackets = await link.endA.querySentPackets();
@@ -319,9 +320,13 @@ test.serial(
     // const totalSent = amounts.reduce((a, b) => a + b, 0);
 
     // let's make 3 transfer tx at different heights on each channel pair
+    interface Meta {
+      height: number;
+      channelId: string;
+    }
     const txHeights = {
-      channels1: [] as number[],
-      channels2: [] as number[],
+      channels1: [] as Meta[],
+      channels2: [] as Meta[],
     };
 
     for (const amount of amounts) {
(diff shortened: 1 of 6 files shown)


src/lib/link.spec.ts
```diff
@@ -285,3 +285,120 @@ test.serial('submit multiple tx, get unreceived packets', async (t) => {
...
+        recipient,
+        destHeight
+      );
+      txHeights.channels2.push(height);
+    }
+
+    // need to wait briefly for it to be indexed
+    await sleep(100);
+
+    // now query for all packets
```
ethanfrey, line 359: Nice thorough tests (here til bottom)

src/lib/link.ts
```diff
@@ -362,32 +362,57 @@ export class Link {
...
+        };
+      },
+      {}
     );
-    const unreceived = new Set(sequences.map((seq) => seq.toNumber()));
-    return allPackets.filter(({ packet }) =>
-      unreceived.has(packet.sequence.toNumber())
+    const unreceivedResponses = await Promise.all(
+      Object.entries(packetsPerDestination).map(([destination, sequences]) => {
+        const [port, channel] = destination.split(':');
```
ethanfrey, line 387: Why split this here?

src/lib/link.ts
```diff
@@ -362,32 +362,57 @@ export class Link {
...
+            channel,
+            sequences
+          ),
+        ]);
+      })
+    );
+
+    const unreceived = unreceivedResponses.reduce(
+      (nested: Record<string, Set<number>>, [port, channel, { sequences }]) => {
+        const key = `${port}${delim}${channel}`;
```
ethanfrey, line 402: When we just recombine it here?

src/lib/link.ts
```diff
@@ -362,32 +362,57 @@ export class Link {
...
+      },
+      {}
     );
-    const unreceived = new Set(sequences.map((seq) => seq.toNumber()));
-    return allPackets.filter(({ packet }) =>
-      unreceived.has(packet.sequence.toNumber())
+    const unreceivedResponses = await Promise.all(
+      Object.entries(packetsPerDestination).map(([destination, sequences]) => {
+        const [port, channel] = destination.split(':');
+        return Promise.all([
```
ethanfrey, line 388: Why we need Promise.all here? There is only one promise.

ethanfrey (approved):
It looks good. The code makes much more sense than before. I think it could be simplified a little further, but happy to merge as is

PR #83, opened 2021-02-25
Feature/55 init

### LINKED ISSUE #55: Add ibc-setup init

### LINKED ISSUE #75: Validate registry.yaml and assert a type
Validate the contents of the `registry.yaml` file after loading it and cast it to useful type.

The current type is too broad:
image

bartmacbartek:
Closes #55 
Closes #75

### DIFF
--- a/src/binary/ibc-setup/commands/init.ts
+++ b/src/binary/ibc-setup/commands/init.ts
@@ -110,7 +110,7 @@ export async function run(options: Options) {
       chains: {
         type: 'object',
         minProperties: 2,
-        required: [options.src, options.dest], // ensure registry compatibility with "src" and "dest"
+        required: [],
         additionalProperties: false,
         patternProperties: {
           '^(.*)$': {
@@ -139,8 +139,17 @@ export async function run(options: Options) {
     );
   }
 
-  const chainSrc = registry.chains[options.src];
-  const chainDest = registry.chains[options.dest];
+  const [chainSrc, chainDest] = [options.src, options.dest].map((chain) => {
+    const chainData = registry.chains[chain];
+
+    if (!chainData) {
+      throw new Error(
+        `Chain ${chain} is missing in the registry, either check the spelling or add the chain definition to ${registryFilePath}`
+      );
+    }
+
+    return chainData;
+  });
 
   const mnemonic = generateMnemonic();
 
--- a/src/binary/ibc-setup/commands/init.ts
+++ b/src/binary/ibc-setup/commands/init.ts
@@ -122,9 +122,8 @@ export async function run(options: Options) {
 
   const [addressSrc, addressDest] = await Promise.all([
     deriveAddress(mnemonic, chainSrc.prefix, chainSrc.hd_path),
-    deriveAddress(mnemonic, chainDest.prefix, chainSrc.hd_path),
+    deriveAddress(mnemonic, chainDest.prefix, chainDest.hd_path),
   ]);
-
   console.log(`Source address: ${addressSrc}`);
   console.log(`Destination address: ${addressDest}`);
 }
--- a/src/binary/ibc-setup/commands/init.ts
+++ b/src/binary/ibc-setup/commands/init.ts
@@ -1,6 +1,7 @@
 import fs from 'fs';
 import path from 'path';
 
+import { Bip39, Random } from '@cosmjs/crypto';
 import axios from 'axios';
 import yaml from 'js-yaml';
 
@@ -13,6 +14,11 @@ export type Options = GlobalOptions & {
 };
 
 const registryFile = 'registry.yaml';
+const appFile = 'app.yaml';
+
+export function generateMnemonic(): string {
+  return Bip39.encode(Random.getBytes(16)).toString();
+}
 
 export function init(flags: Partial<Options>) {
   function getDefaultHome() {
@@ -33,6 +39,13 @@ export function init(flags: Partial<Options>) {
 }
 
 export async function run(options: Options) {
+  console.log(generateMnemonic());
+  const appFilePath = path.join(options.home, appFile);
+  if (fs.existsSync(appFilePath)) {
+    console.log(`The ${appFile} is already initialized at ${options.home}`);
+    return;
+  }
+
   if (!fs.existsSync(options.home)) {
     fs.mkdirSync(options.home, { recursive: true });
     console.log(`Initialized home directory at ${options.home}`);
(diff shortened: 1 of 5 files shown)


src/binary/ibc-setup/commands/init.ts
```diff
@@ -12,7 +14,41 @@ export type Options = GlobalOptions & {
   readonly dest: string;
 };
 
+type Registry = {
+  version: string;
+  chains: Record<
+    string,
+    {
```
ethanfrey, line 21: Can you declare a separate (non-inline) type for this? eg. `chains: Record<string, ChainInfo>` and then that last part as ChainInfo (I assume we will just want that in many places)

src/binary/ibc-setup/commands/init.ts
```diff
@@ -33,6 +69,12 @@ export function init(flags: Partial<Options>) {
 }
 
 export async function run(options: Options) {
+  const appFilePath = path.join(options.home, appFile);
+  if (fs.existsSync(appFilePath)) {
+    console.log(`The ${appFile} is already initialized at ${options.home}`);
```
ethanfrey, line 74: :+1:

src/binary/ibc-setup/commands/init.ts
```diff
@@ -46,17 +88,42 @@ export async function run(options: Options) {
...
+  const mnemonic = generateMnemonic();
+
+  const appYaml = yaml.dump(
+    {
+      src: options.src,
+      dest: options.dest,
+      mnemonic,
+    },
+    {
+      lineWidth: 1000,
```
ethanfrey, line 116: Ah, nice to avoid the break. 1000 = never line break

Can you add a comment ("to ensure mnemonic is not split on multiple lines")

src/binary/ibc-setup/commands/init.ts
```diff
@@ -46,17 +88,42 @@ export async function run(options: Options) {
...
   }
 
-  const registry = yaml.load(fs.readFileSync(registryFilePath, 'utf-8'));
+  const registry = yaml.load(
+    fs.readFileSync(registryFilePath, 'utf-8')
+  ) as Registry;
   // TODO #75: registry file validation
-  console.log(registry);
+
+  const chainSrc = registry.chains[options.src];
```
ethanfrey, line 104: These could be undefined, right?
We should validate that both `chainSrc` and `chainDest` are objects here, before making the `app.yaml` file. (Try to `--src=foobar` and see what happens.)

src/binary/ibc-setup/commands/init.ts
```diff
@@ -46,17 +88,42 @@ export async function run(options: Options) {
...
+      dest: options.dest,
+      mnemonic,
+    },
+    {
+      lineWidth: 1000,
+    }
+  );
+
+  fs.writeFileSync(appFilePath, appYaml, { encoding: 'utf-8' });
+  console.log(`Saved configuration to ${appFilePath}`);
```
ethanfrey, line 121: We will be using a logger here soon, but yes, console.log is good for now

ethanfrey (commented):
Nice work.

A few minor comments on error handling, but with a little polish, looks good to merge

src/binary/ibc-setup/commands/init.ts
bartmacbartek: 51b003c

src/binary/ibc-setup/commands/init.ts
bartmacbartek: 117f41f

src/binary/ibc-setup/commands/init.ts
bartmacbartek: Yeah, I was going to cover it with registry validation.

- aefe0ec
- ecf1796

src/binary/ibc-setup/commands/init.ts
```diff
@@ -57,6 +97,71 @@ export async function run(options: Options) {
...
+    },
+  };
+  const validate = ajv.compile(schema);
+  if (!validate(registry)) {
+    const errors = (validate.errors ?? []).map(
+      ({ dataPath, message }) => `"${dataPath}" ${message}`
+    );
+    throw new Error(
+      [`${registryFile} validation failed.`, ...errors].join(os.EOL)
+    );
```
bartmacbartek, line 139: It produces this kind of errors:

```bash
Error: registry.yaml validation failed.
"/chains/local_wasm" should have required property 'chain_id'
```

src/binary/ibc-setup/commands/init.ts
```diff
@@ -57,6 +97,71 @@ export async function run(options: Options) {
   }
 
   const registry = yaml.load(fs.readFileSync(registryFilePath, 'utf-8'));
-  // TODO #75: registry file validation
-  console.log(registry);
+
+  const ajv = new Ajv({ allErrors: true });
+  const schema: JSONSchemaType<Registry> = {
```
ethanfrey, line 102: Nice one.
I'd love to put this in validation logic, and maybe load load and validate, in another file (but as you said, that can be next PR)

src/binary/ibc-setup/commands/init.ts
ethanfrey: Yeah, looks nice.
When we add a logger, we can look if there is a better format, but it looks nice (and nice use of join)

src/binary/ibc-setup/commands/init.ts
ethanfrey: THe registry is valid, but you don't check if these keys are present, correct?

Seems like 3 lines to add. This is the only blocker before merge

ethanfrey (approved):
Looks good, please just ensure it throws if I pass --src or --dest not in the registry

src/binary/ibc-setup/commands/init.ts
bartmacbartek: It's already there: https://github.com/confio/ts-relayer/pull/83/files#diff-e8c0ad1a88c4869847d33b7631f1c6a11037972cd3ac65ecab3834ef0f0bf617R113

src/binary/ibc-setup/commands/init.ts
bartmacbartek: Of course, I think we would have to revisit all the error messages sooner or later. It'd be nice to have them in the same style and format.

src/binary/ibc-setup/commands/init.ts
bartmacbartek: Yes, especially that some functionality must be shared across the commands. I think it's better to start working on this while having at least 2 usages.

src/binary/ibc-setup/commands/init.ts
```diff
@@ -110,7 +110,7 @@ export async function run(options: Options) {
       chains: {
         type: 'object',
         minProperties: 2,
-        required: [options.src, options.dest], // ensure registry compatibility with "src" and "dest"
+        required: [],
```
bartmacbartek, line 113: The `required` property can't be removed due to schema type.

Issue #52, opened 2021-02-18
Basic key management commands


ethanfrey:
Builds on #51 

* Use [key configuration flags from spec](https://github.com/confio/ts-relayer/blob/main/spec/keys.md) - if `-i` is hard, that can be done as a later issue
* Implement `ibc-setup` [commands for key management](https://github.com/confio/ts-relayer/blob/main/spec/ibc-setup.md#account-management)
* Test this in some automated test - can use a test config with only CI, and hit the faucets to get non-zero balances

ethanfrey:
Closed by #85 and #91

PR #85, opened 2021-03-01
52 - basic key management commands

bartmacbartek:
addresses #52


src/binary/ibc-setup/utils/resolve-option.ts
```diff
@@ -1,14 +1,13 @@
 export function resolveOption(
-  identifier: string,
-  ...args: Array<(string | undefined) | (() => string | undefined | void)>
+  ...args: Array<(string | undefined) | (() => string | undefined)>
```
bartmacbartek, line 2: It turned out that `void` adds unnecessary complexity to handling. I removed it as you requested before. @ethanfrey

src/binary/ibc-setup/utils/resolve-required-option.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+test('uses resolveOption func', (t) => {
+  const option = resolveRequiredOption(
+    'some option', // identifier
+    undefined,
+    undefined,
+    'some string',
+    'another string'
+  );
+  t.is(option, 'some string');
+});
```
bartmacbartek, line 14: Since `resolveRequiredOption` is a proxy to `resolveOption`, I put this test here to make sure that everything's okay.

src/binary/ibc-setup/commands/keys-generate.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+const consoleLog = sinon.stub(console, 'log');
+const fsWriteFileSync = sinon.stub(fs, 'writeFileSync');
+
+test.beforeEach(() => {
+  sinon.reset();
+
+  fsWriteFileSync.returns();
+});
+
+test('generates mnemonic to stdout', (t) => {
```
ethanfrey, line 17: Nice tests :slightly_smiling_face:

src/binary/ibc-setup/commands/keys-generate.ts
```diff
@@ -0,0 +1,32 @@
...
+import { generateMnemonic } from '../utils/generate-mnemonic';
+import { resolveOption } from '../utils/resolve-option';
+
+export type Options = {
+  keyFile?: string;
+};
+
+export function keysGenerate(flags: Options) {
+  const options = {
+    keyFile: resolveOption(
```
ethanfrey, line 12: Nice to have the required/optional split here

src/binary/ibc-setup/commands/keys-generate.ts
```diff
@@ -0,0 +1,32 @@
...
+      'key-file',
```
ethanfrey, line 13: `resolveOption` no longer takes an identifier as the first arg (properly), so this should be removed (now it means the option value will always be `key-file`, correct?).

Maybe to avoid such a mistake in the future, you could do something like `requireOption('key-file')(flags.keyFile, process.env.RELAYER_KEY_FILE)` easier to remove the extra call than a compatible arg when doing a refactor

ethanfrey (commented):
Nice stuff and good start.
Please test this once manually (the resolveOption issue mentioned below).

Do you need sample js code for the other two commands? I guess listing keys you have already. The listing of balances, I will show in a PR

src/binary/ibc-setup/commands/keys-generate.ts
bartmacbartek: Indeed, this is a bug. `resolveOption` doesn't take the identifier anymore but I curried the `resolveRequiredOption`: 45a6278

src/binary/ibc-setup/commands/init.ts
```diff
@@ -96,48 +59,7 @@ export async function run(options: Options) {
...
-  const validate = ajv.compile(schema);
-  if (!validate(registry)) {
-    const errors = (validate.errors ?? []).map(
-      ({ dataPath, message }) => `"${dataPath}" ${message}`
-    );
-    throw new Error(
-      [`${registryFile} validation failed.`, ...errors].join(os.EOL)
-    );
-  }
+  const registry = loadAndValidateRegistry(registryFilePath);
```
ethanfrey, line 62: :+1:

src/binary/ibc-setup/commands/keys-list.ts
```diff
@@ -0,0 +1,93 @@
...
+  readonly keyFile?: string;
+  readonly mnemonic?: string;
+};
+
+export type Options = {
+  readonly home: string;
+  readonly mnemonic: string;
+};
+
+export async function readMnemonicFromStdin(interactive: Flags['interactive']) {
```
ethanfrey, line 24: Oh, nice, you added the interactive support.

src/binary/ibc-setup/commands/keys-list.ts
```diff
@@ -0,0 +1,93 @@
...
+    readlineInterface.question('enter mnemonic phrase: ', (stdin) => {
+      readlineInterface.close();
+      resolve(stdin);
+    });
+  });
+
+  return mnemonic;
+}
+
+function readMnemonicFromFile(keyFile: Flags['keyFile']) {
```
ethanfrey, line 44: I would pull both these key file functions into utils. We need them as well when listing balances as well. Maybe add some flag parsing/handling logic into that utils as well, as it is the same for all commands that use it (the second command to use it is in the same issue)

src/binary/ibc-setup/commands/keys-list.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export async function keysList(flags: Flags) {
+  const options: Options = {
+    home: resolveRequiredOption('home')(
+      flags.home,
+      process.env.RELAYER_HOME,
+      getDefaultHomePath
+    ),
+    mnemonic: resolveRequiredOption('mnemonic')(
```
ethanfrey, line 61: This is a very interesting way to handle this. It surprised me at first and I was like "no, this doesn't make sense". But seeing how each of these (readMnemonicFromStdin and readMnemonicFromFile) returns undefined if the flag was not set (becomes a no-op), this is a great way to handle order of operations:

* if `-i`, then read from stdin
* if `--mnemonic`, use it
* if `--keyfile`, load from file
* if `RELAYER_MNEMONIC` set, use that

My only open question now is how the `app.yaml` flag values are loaded?

src/binary/ibc-setup/utils/resolve-required-option.ts
```diff
@@ -0,0 +1,13 @@
+import { resolveOption } from './resolve-option';
+
+export function resolveRequiredOption(identifier: string) {
```
ethanfrey, line 3: :+1:

ethanfrey (commented):
This does not fully close the issue, but if you change the PR title to "addresses #52", I think this is ready to merge and then can finish the last command in another PR.

And the app.yaml parsing

ethanfrey:
Hi level (usage) review:

```
ibc-setup init --src local_wasm --dest local_simapp
ibc-setup keys list
```

Returns error (doesn't use app.yaml as default).

```
ibc-setup keys list --mnemonic="$(ibc-setup keys generate)"
```

Works as expected, as does

```
ibc-setup keys generate --key-file=foo.key
ibc-setup keys list --key-file=foo.key
```

src/binary/ibc-setup/commands/keys-list.ts
ethanfrey: After more reflection I think the proper handling would be to check each flag one after another. Something like:

```ts
let mnemonic = await readMnemonicFromStdin(flags.mneomic);
if (!mnemonic) {
  const mnemonic = resolveOption(flags.mnemonic, app.mnemonic, process.env.RELAY_MNEMONIC);
}
if (!mnemonic) {
  const keyfile = resolveOption(flags.keyFile, app.keyFile, process.env.RELAY_KEY_FILE);
  if (keyfile) {
    mnemonic = readMnemonicFromFile(keyfile);
  }
}
if (!mnemonic) {
  throw new Error('You must provide the mnemonic via `interactive`, `mnemonic` or `key-file`');
```

ethanfrey: Yes, this is not so clear as using the same loop, but the logic is different. We can wrap this in one helper function that takes flags and app config. and just use it.

The `resolveOption` and `requireResolveOption` can be used everywhere else, but not for doing the checks over multiple possible flags

bartmacbartek: I see your point, but I wanted to give it one more try: eb4a75e

Here, we prepare the `keyFile` before resolving the mnemonic:
```ts
const home = resolveRequiredOption('home')(...)
const app = loadAndValidateApp(home);
const keyFile = resolveOption(flags.keyFile, app?.keyFile, process.env.KEY_FILE);

const mnemonic = await resolveMnemonicOption({ app, keyFile, interactive: flags.interactive, mnemonic: flags.mnemonic });
```

In **resolveMnemonicOption**, the `resolveRequiredOption` is used to determine the order.

**Note: ** I introduced `null`s to indicate values explicitly set to "nothing" by us. It may look like a mishmash at first, but it does make sense when you think of `undefined` as a value that comes from the outside (flags, configs) and a `null` that is explicitly set in the program. I'm not 100% sure about this, but it gives me more confidence.

What do you think of such an approach (mnemonic and nulls)?

src/binary/ibc-setup/commands/keys-list.ts
bartmacbartek: Sure, this will be reused between commands.

src/binary/ibc-setup/utils/load-and-validate-app.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    if (!validate(app)) {
+      return null;
+    }
+
+    return app;
+  } catch {
+    return null;
+  }
+}
```
bartmacbartek, line 36: The `app.yaml` is optional, thus this function return type is `App | null`. Does it make sense? @ethanfrey

src/binary/ibc-setup/utils/resolve-option.ts
```diff
@@ -1,14 +1,13 @@
 export function resolveOption(
-  identifier: string,
-  ...args: Array<(string | undefined) | (() => string | undefined | void)>
+  ...args: Array<(string | undefined | null) | (() => string | null)>
```
bartmacbartek, line 2: `undefined  | null` because it can accept "nothing" values from outside of the program (flags, configs) and values resolved to `null` in the program.

src/binary/ibc-setup/utils/load-and-validate-app.ts
ethanfrey: Yes.

ethanfrey: I would only catch errors around the missing file (read errors), and return null or undefined if missing.

The validation error should throw, so people know why the values in the file are not being used

bartmacbartek: 00fb34f

src/binary/ibc-setup/utils/load-and-validate-app.ts
```diff
@@ -8,29 +9,38 @@ import { appFile } from '../../constants';
...
-
-    if (!validate(app)) {
-      return null;
-    }
-
-    return app;
-  } catch {
+  const appFilePath = path.join(home, appFile);
+
+  if (!fs.existsSync(appFilePath)) {
```
ethanfrey, line 14: Nice

src/binary/ibc-setup/utils/load-and-validate-app.ts
```diff
@@ -8,29 +9,38 @@ import { appFile } from '../../constants';
...
+    },
+  };
+  const validate = ajv.compile(schema);
+
+  if (!validate(app)) {
+    const errors = (validate.errors ?? []).map(
+      ({ dataPath, message }) => `"${dataPath}" ${message}`
+    );
+    throw new Error([`${appFile} validation failed.`, ...errors].join(os.EOL));
+  }
```
bartmacbartek, line 43: This is copied-pasted but we're going to clean it up after implementing the logger.

willclarktech (commented):
A few comments

src/binary/ibc-setup/utils/get-default-home-path.spec.ts
```diff
@@ -0,0 +1,26 @@
+import test from 'ava';
+
+import { getDefaultHomePath } from './get-default-home-path';
+
+const processEnvCopy = { ...process.env };
```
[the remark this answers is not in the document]
bartmacbartek: Yes, that's why I make a copy first and then before each test I "reset" it:
```
test.beforeEach(() => {
  process.env = processEnvCopy;
});
```

src/binary/ibc-setup/utils/get-default-home-path.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  process.env = processEnvCopy;
+});
+
+test('returns path if $HOME variable is set', (t) => {
+  process.env.HOME = '/home/user';
+  t.is(getDefaultHomePath(), '/home/user/.ibc-setup');
+
+  process.env.HOME = '/home/pathEndingWithSlash/';
+  t.is(getDefaultHomePath(), '/home/pathEndingWithSlash/.ibc-setup');
+});
```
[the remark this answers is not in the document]
bartmacbartek: Do you mean that by default `$HOME` is not ending with a slash? That's correct, however, someone can do `export HOME=/home/user/` still. IMO it's good to have it covered.

src/binary/ibc-setup/utils/resolve-required-option.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+    null,
+    'some string',
+    'another string'
+  );
+  t.is(option, 'some string');
+});
+
+test('throws if all options are undefined or null', (t) => {
+  const option = () =>
+    resolveRequiredOption('some option')(undefined, null, undefined);
```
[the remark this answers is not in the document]
bartmacbartek: 09875ef

src/binary/ibc-setup/utils/resolve-mnemonic-option.ts
```diff
@@ -0,0 +1,58 @@
...
+  interactive,
+  keyFile,
+  mnemonic: mnemonicFlag,
+  app,
+}: Params) {
+  return resolveRequiredOption('mnemonic')(
+    await readMnemonicFromStdin(interactive),
+    mnemonicFlag,
+    app?.mnemonic,
+    process.env.RELAYER_MNEMONIC,
```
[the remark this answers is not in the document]
bartmacbartek: You're right, thanks!

>there is an optional app.yaml file that is used as a fall back for any environmental variable. 

0f4f9df

src/binary/ibc-setup/utils/load-and-validate-app.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function loadAndValidateApp(home: string) {
+  const appFilePath = path.join(home, appFile);
+
+  if (!fs.existsSync(appFilePath)) {
+    return null;
+  }
+
+  if (!fs.lstatSync(appFilePath).isFile()) {
+    throw new Error(`${appendFileSync} must be a file.`);
```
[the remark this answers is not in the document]
bartmacbartek: 🤯 

fixed in 8ae637e

src/binary/ibc-setup/utils/load-and-validate-app.ts
```diff
@@ -0,0 +1,46 @@
...
+import Ajv, { JSONSchemaType } from 'ajv';
+import yaml from 'js-yaml';
+
+import { appFile } from '../../constants';
+import { App } from '../types';
+
+export function loadAndValidateApp(home: string) {
+  const appFilePath = path.join(home, appFile);
+
+  if (!fs.existsSync(appFilePath)) {
```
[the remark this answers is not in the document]
bartmacbartek: Good point about race conditions.

Does it look better?
8ae637e

I'm thinking if we should try..catch the `yaml.load` (both in app and registry loading)? WYT?

src/binary/ibc-setup/utils/resolve-mnemonic-option.ts
```diff
@@ -0,0 +1,58 @@
...
+import { resolveRequiredOption } from './resolve-required-option';
+
+async function readMnemonicFromStdin(interactive: boolean) {
+  if (!interactive) {
+    return null;
+  }
+
+  const readlineInterface = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
```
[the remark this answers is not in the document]
bartmacbartek: Yes, w/o output it doesn't work:
>The output stream is used to print prompts for user input that arrives on, and is read from, the input stream.

src/binary/ibc-setup/utils/resolve-mnemonic-option.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  const mnemonic = await new Promise<string>((resolve) => {
+    readlineInterface.question('enter mnemonic phrase: ', (stdin) => {
+      readlineInterface.close();
+      resolve(stdin);
```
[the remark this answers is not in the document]
bartmacbartek: here: 7fb1824

Should we catch this rejection or leave it for later (the error boundary)?

src/binary/ibc-setup/types.ts
```diff
@@ -1,3 +1,19 @@
...
+  hd_path: string;
+  rpc: string[];
+};
+
+export type Registry = {
+  version: number;
+  chains: Record<string, Chain>;
+};
+
+export type App = {
```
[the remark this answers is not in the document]
bartmacbartek: `AppConfig` ✅ 
aa78df4

src/binary/ibc-setup/utils/resolve-option.ts
bartmacbartek:
```ts
const keyFile = resolveOption(flags.keyFile, process.env.KEY_FILE, app?.keyFile); // can be null
const mnemonic = resolveRequiredOption('mnemonic')(keyFile, ...);
```

src/binary/ibc-setup/utils/get-default-home-path.spec.ts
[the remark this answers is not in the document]
willclarktech: Right, but what about after the last test?

src/binary/ibc-setup/utils/get-default-home-path.spec.ts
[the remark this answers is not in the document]
willclarktech: I mean, how about testing a case where `$HOME` doesn't *start* with a slash. Eg `~` or something totally unexpected like `"i messed up my $HOME env variable somehow"`.

src/binary/ibc-setup/utils/resolve-mnemonic-option.ts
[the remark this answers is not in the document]
willclarktech: I would think this can wait until we have a general approach for error handling.

src/binary/ibc-setup/utils/load-and-validate-app.ts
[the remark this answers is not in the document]
willclarktech: I think if `fs.readFileSync` throws we don't want to silently swallow that error. It probably means the user intended to use a config file but it doesn't have the right permissions or they gave us the enclosing folder instead of the file or something.

src/binary/ibc-setup/utils/get-default-home-path.spec.ts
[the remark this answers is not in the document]
bartmacbartek: No need to reset it after the last test because
>AVA is highly opinionated and runs tests concurrently, with a separate process for each test file.

Or do I understand it wrong?

src/binary/ibc-setup/utils/resolve-mnemonic-option.ts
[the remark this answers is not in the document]
bartmacbartek: Good, resolving.

src/binary/ibc-setup/utils/load-and-validate-app.ts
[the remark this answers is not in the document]
bartmacbartek: The thing is that `app.yaml` is optional so it may not-exist. In such case we'd like to return `null` from `loadAndValidateApp`

src/binary/ibc-setup/utils/get-default-home-path.spec.ts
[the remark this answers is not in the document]
willclarktech: Ah great, thanks ava.

src/binary/ibc-setup/utils/load-and-validate-app.ts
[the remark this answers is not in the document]
willclarktech: Yeah, but you'll get different errors for different situations. I think there are at least a couple where we want to exit immediately and tell the user they're doing something wrong.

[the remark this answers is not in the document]
ethanfrey: I think the original logic worked relatively well. It silently ignored missing app.yaml and threw errors if it was a directory or we didn't have read permissions.

There is a race condition, but that is a super edge case. Just make an issue for that. We are going for 0.1.0 release. Let's try to get "good enough" out and add issues for ways to improve it in the future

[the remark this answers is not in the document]
ethanfrey: I don't have a strong preference over either of the two approaches done. Both have limits. Let's mark it, merge and move forward. Merging this blocks writing more commands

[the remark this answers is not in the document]
bartmacbartek: I pushed something that eliminates race condition and returns `null` if the file doesn't exist. It throws on other errors though such as no permissions.

a66467d

src/binary/ibc-setup/utils/get-default-home-path.spec.ts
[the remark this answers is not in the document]
bartmacbartek: Yeah, we don't validate the path at this stage. In the case of `i messed up $HOME env variable` the user should be notified later at some `fs.readFile` step.

Maybe we should validate the path but let's do it in a follow-up and merge this now.

ethanfrey (approved):
Good to merge once it passes CI

PR #87, opened 2021-03-01
79 support nonzero revision number

### LINKED ISSUE #79: Support non-zero RevisionNumber in IBC packets
I assume this is 0 everywhere in the code and it worked in the CI. I wondered where/when it would be set. This came up recently on the cosmos discord, and here was the resolution:

```
Doing some testing here with relayer debug turned on, I think the above statement has it in the wrong direction.  The last statement should read: IBC packets destined to the chain (as opposed to originating from) must have their "revision_number" field set to the chain id suffix.

In testing, I have two chains: "gaiad-microtick-testnet" which is revision number = 0, and "microtick-testnet-rc3-1" which would be revision number = 1.  In order for the packet commitments to verify correctly, IBC packets from gaiad -> microtick need the revision number set to "1".  In the other direction, packets from microtick -> gaiad need the revision number set to "0".

Using these values, everything seems to be working correctly again.  I wanted to post the solution here in case anyone else comes across this.

Thanks for your help guys!
```

So, it is based on the chain-id, and we will need that to connect to most networks. It is easy to add a suffix in the CI chains to enforce this. Then we need to ensure support throughout the stack until all tests pass again.

ethanfrey:
Closes #79 

* Regenerated CI blockchains to demo the problem (wasmd has revisionNumber = 1)
* Removed default (0) revisionNumber in all locations I found
* All files compiles

Currently, there is an issue with the light client proofs when trying to prove a new connection. First using the wrong revision number, now getting a non-existence proog


scripts/wasmd/generate_template.sh
```diff
@@ -29,37 +32,27 @@ if [ ! -x "$SCRIPT_DIR/template/.wasmd/config/gentx" ]; then
...
-    -e 's/^timeout_precommit =.*$/timeout_precommit = "300ms"/' \
+    -e 's/^timeout_precommit =.*$/timeout_precommit = "100ms"/' \
     -e 's/^timeout_precommit_delta =.*$/timeout_precommit_delta = "100ms"/' \
-    -e 's/^timeout_commit =.*$/timeout_commit = "1s"/' \
+    -e 's/^timeout_commit =.*$/timeout_commit = "200ms"/' \
     "template/.wasmd/config/config.toml"
 
   # Custom settings app.toml
-  sed -i "" \
+  sed -i"" \
```
willclarktech, line 55: Are these from `shfmt`?

scripts/wasmd/template/.wasmd/config/config.toml
```diff
@@ -158,8 +158,8 @@ max_header_bytes = 1048576
 tls_cert_file = ""
 
 # The path to a file containing matching private key that is used to create the HTTPS server.
-# Migth be either absolute path or path related to tendermint's config directory.
-# NOTE: both tls_cert_file and tls_key_file must be present for Tendermint to create HTTPS server.
+# Might be either absolute path or path related to Tendermint's config directory.
+# NOTE: both tls-cert-file and tls-key-file must be present for Tendermint to create HTTPS server.
```
willclarktech, line 162: Was this change from us? It seems incorrect here.

scripts/wasmd/template/.wasmd/config/config.toml
```diff
@@ -259,13 +259,19 @@ max_txs_bytes = 1073741824
 # Size of the cache (used to filter transactions we saw earlier) in transactions
 cache_size = 10000
 
+# Do not remove invalid transactions from the cache (default: false)
+# Set to true if it's not possible for any invalid transaction to become valid
+# again in the future.
+keep-invalid-txs-in-cache = false
```
willclarktech, line 265: How come this isn't snake_case?

src/lib/ibcclient.ts
```diff
@@ -257,6 +272,27 @@ export class IbcClient {
...
   }
 
+  public revisionHeight(height: number): Height {
+    return Height.fromPartial({
+      revisionHeight: new Long(height),
+      revisionNumber: this.revisionNumber,
+    });
+  }
+
+  public ensureRevisionHeight(height: number | Height): Height {
```
willclarktech, line 282: What if it's a `Height` but the `revisionNumber`s don't match?

src/lib/queries/ibc.ts
```diff
@@ -684,21 +687,18 @@ export function setupIbcExtension(base: QueryClient): IbcExtension {
           },
           consensusState: async (
             clientId: string,
-            consensusHeight: number,
-            proveHeight?: number
+            consensusHeight: Height,
+            proofHeight: Height
           ) => {
-            // TODO: accept other revisionNumber
```
willclarktech, line 690: 👍

src/lib/utils.spec.ts
```diff
@@ -0,0 +1,17 @@
+import test from 'ava';
+
+import { parseRevisionNumber } from './utils';
+
+test('can parse revision numbers', (t) => {
+  const musselnet = parseRevisionNumber('musselnet-4');
+  t.is(musselnet.toNumber(), 4);
+
+  const numerific = parseRevisionNumber('numers-123-456');
```
willclarktech, line 9: Should this throw? If not how about these test cases?
- `''`
- `'-'`
- `'hello-'`
- `'hello-123-'`
- `'hello-0123'`
- `'hello-00123'`
- `'hello-1.23'`

scripts/wasmd/generate_template.sh
ethanfrey: No. The previous way only works on osx and not linux.

I had this issue before. Linux wants `sed -i 's/...'`. OSX wants `sed -i "" 's/...'`. The only way that works on both is `sed -i"" 's/...'`

scripts/wasmd/template/.wasmd/config/config.toml
ethanfrey: Autogenerated by wasmd / cosmos-sdk

scripts/wasmd/template/.wasmd/config/config.toml
ethanfrey: Autogenerated by wasmd / cosmos-sdk (like all the config changes)

src/lib/ibcclient.ts
ethanfrey: I guess I can assert that as well. Good point

src/lib/utils.spec.ts
ethanfrey: It should never throw. This is the description of what the SDK does (from cosmos discord)

```
So currently it looks like any chain ID with an ending "-", followed by a non-zero digit, then any number of [0-9] digits will be interpreted as being in "Epoch Format", so for those chains, IBC packets originating from that chain must have their "revision_number" field set to that number.
```

ethanfrey:
@willclarktech Addressed all PR issues. Can you give final review?

src/lib/ibcclient.ts
```diff
@@ -257,6 +272,34 @@ export class IbcClient {
...
+    } else {
+      if (height.revisionNumber.toNumber() !== revisionNumber.toNumber()) {
+        throw new Error(
+          `Using incorrect revisionNumber ${height.revisionNumber} on chain with ${this.revisionNumber}`
+        );
+      }
+      revisionHeight = height.revisionHeight;
+    }
+    return Height.fromPartial({ revisionHeight, revisionNumber });
+  }
```
willclarktech, line 296: This seems simpler?
```ts
if (typeof height === 'number') {
  return Height.fromPartial({ revisionHeight: new Long(height), revisionNumber: this.revisionNumber });
}
if (height.revisionNumber.toNumber() !== this.revisionNumber.toNumber()) {
  throw new Error("xxx");
}
return height;
```

willclarktech (commented):
Looking good, just one suggestion.

src/lib/ibcclient.ts
ethanfrey: Yes, simpler.

Note: If you are not going to be at the computer and responsive and it is a small, clear change, can you just approve it? I will merge anyway (admin rights), but we have a pile up of PRs and I think a bit of pre-emptive approvals (when you can trust the dev to just implement the minor change requests) is a better flow.

PR #90, opened 2021-03-02
Implement timeouts

### LINKED ISSUE #45: Implement Timeouts Submission
After #43 and #44

If a packet is pending, but it is too late to submit it, we should relay a timeout to the original chain (with a non-existence proof from the intended recipient). This is similar to submitting acks, but the queries are a bit different.

willclarktech:
Closes #45


src/lib/utils.ts
```diff
@@ -289,3 +291,33 @@ export function multiplyCoin({ amount, denom }: Coin, mult: number): Coin {
   const multAmount = Number.parseInt(amount, 10) * mult;
   return { amount: multAmount.toString(), denom };
 }
+
+export function splitPendingPackets(
+  currentHeight: number,
```
ethanfrey, line 296: you need to take timeoutTimestamp as well (measured in nanoseconds)

src/lib/link.ts
```diff
@@ -520,6 +521,39 @@ export class Link {
...
+    this.logger.info(`Timeout packets for source ${source}`);
+    console.log('N PACKETS:', packets.length);
+    const { src, dest } = this.getEnds(source);
+
+    // check if we need to update client at all
+    const neededHeight = Math.max(...packets.map((x) => x.height)) + 1;
+    // const headerHeight = await this.updateClientToHeight(source, neededHeight);
+    await this.updateClientToHeight(source, neededHeight);
+
+    const [packet] = packets;
```
ethanfrey, line 537: Obviously you will need a loop here, but you know that.

Probably just update client.timeoutPacket -> client.timeoutPackets that takes an array of packets with proofs

src/lib/link.ts
```diff
@@ -520,6 +521,39 @@ export class Link {
...
+    const nextSequenceRecvResponse = await dest.client.getNextSequenceRecvProof(
```
ethanfrey, line 538: You need the same proof query as for submitting an ack. However, when you get it the value should be empty and the proof existing (it is a non-existence proof) - you can console.log the query to ensure this works

src/lib/link.ts
```diff
@@ -520,6 +521,39 @@ export class Link {
     return height;
   }
 
+  public async timeoutPackets(
+    source: Side,
+    packets: readonly PacketWithMetadata[]
+  ): Promise<readonly MsgResult[]> {
```
ethanfrey, line 527: I wouldn't really return all results. It fails if the tx fails.
Maybe return the block height where they were submitted? Or nothing until you find a need

src/lib/link.spec.ts
```diff
@@ -484,3 +485,85 @@ test.serial(
...
+  // let's make 3 transfer tx at different heights
+  const txHeights = [];
+  for (let i = 0; i < amounts.length; ++i) {
+    const token = { amount: amounts[i].toString(), denom: simapp.denomFee };
+    const { height } = await nodeA.transferTokens(
+      channels.src.portId,
+      channels.src.channelId,
+      token,
+      recipient,
+      timeoutHeights[i]
```
ethanfrey, line 526: It would be good to make:

1. expired via height
2. non-expired
3. expired via time

src/lib/link.spec.ts
```diff
@@ -484,3 +485,85 @@ test.serial(
...
+  const noPackets = await link.endA.querySentPackets();
+  t.is(noPackets.length, 0);
+
+  // some basic setup for the transfers
+  const recipient = randomAddress(wasmd.prefix);
+  const latestHeight = (await nodeB.latestHeader()).height;
+  const timeoutDestHeight = latestHeight + 2;
+  const submitDestHeight = latestHeight + 500; // valid for 500 blocks
+  const amounts = [1000, 2222, 3456];
+  const timeoutHeights = [
```
ethanfrey, line 511: This is good for now to ensure splitting works

ethanfrey (commented):
Good start, just a few tweaks and this should work

src/lib/link.ts
willclarktech: Great, I originally returned nothing here, and then switched to this but it's premature optimisation.

PR #91, opened 2021-03-02
52 balances command

bartmacbartek:


src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const options: Options = {
+    home,
+    mnemonic: await resolveMnemonicOption({
+      interactive: flags.interactive,
+      mnemonic: flags.mnemonic,
+      keyFile: keyFile,
+      app,
+    }),
+  };
```
bartmacbartek, line 40: This will get extracted and shared.

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,74 @@
...
+export async function run(options: Options) {
+  const registryFilePath = path.join(options.home, registryFile);
+  const registry = loadAndValidateRegistry(registryFilePath);
+
+  const addresses = await getAddresses(registry.chains, options.mnemonic);
+
+  for (const [chain, data, address] of addresses) {
+    const signer = await DirectSecp256k1HdWallet.fromMnemonic(
+      options.mnemonic,
+      undefined,
```
bartmacbartek, line 54: Should we provide `hd_path` here?

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,74 @@
...
+      data.rpc[0], // rpc[0] is guaranteed to be defined by registry validator
+      signer,
+      address,
+      {
+        prefix: data.prefix,
+        gasPrice,
+      }
+    );
+
+    const balances = await client.query.bank.unverified.allBalances(address);
```
bartmacbartek, line 70: It returns `Coin[]` type.

Should it be iterated and summed?

---

The spec for the `balances` command says:
>prints out all chains for which the relayer has a non-zero balance

does it mean that we need at least one non-zero entry in the `Coin[]` array to include the address in the output?

src/binary/ibc-setup/commands/keys-list.ts
```diff
@@ -49,24 +49,27 @@ export async function keysList(flags: Flags) {
...
       chains.map(([, data]) =>
-        deriveAddress(options.mnemonic, data.prefix, data.hd_path)
+        deriveAddress(mnemonic, data.prefix, data.hd_path)
       )
     )
-  )
-    .map((address, index) => {
-      const chain = chains[index][0];
+  ).map((address, index) => [chains[index][0], chains[index][1], address]);
+}
```
bartmacbartek, line 65: I think I'll lift this one up.

src/binary/ibc-setup/utils/load-and-validate-registry.ts
```diff
@@ -34,7 +34,7 @@ export function loadAndValidateRegistry(filepath: string) {
               prefix: { type: 'string' },
               gas_price: { type: 'string' },
               hd_path: { type: 'string' },
-              rpc: { type: 'array', items: { type: 'string' } },
+              rpc: { type: 'array', items: { type: 'string' }, minItems: 1 },
```
bartmacbartek, line 37: Ensures that the `rpc` field in the registry must contain at least one item.

src/binary/ibc-setup/commands/balances.ts
bartmacbartek: Here: 9c519b7

src/binary/ibc-setup/commands/balances.ts
ethanfrey: Will can you provide a coinToString function for output?

ethanfrey: Maybe we only query the feeDenom, which is the only one that is there

ethanfrey (commented):
Some comments. Nice work

src/binary/ibc-setup/commands/balances.ts
ethanfrey: yes

src/binary/ibc-setup/commands/balances.ts
ethanfrey: there is a `bank.unverified.balance(address, denom)` function

willclarktech: Looks like we don't have one in CosmJS, but you can see the usual format here: https://github.com/cosmos/cosmjs/blob/72b7c0b/packages/launchpad/src/coins.ts#L19

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,86 @@
+import os from 'os';
+import path from 'path';
+
+import { stringToPath } from '@cosmjs/crypto';
+import { GasPrice } from '@cosmjs/launchpad';
+import { DirectSecp256k1HdWallet } from '@cosmjs/proto-signing';
+import { Coin } from '@cosmjs/proto-signing/build/codec/cosmos/base/v1beta1/coin';
```
bartmacbartek, line 7: Looks like the `Coin` is not exported. Should I drop it?

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,86 @@
...
+          address,
+          {
+            prefix: data.prefix,
+            gasPrice,
+          }
+        );
+
+        const balance = await client.query.bank.unverified.balance(
+          address,
+          gasPrice.denom
```
bartmacbartek, line 67: Used `denom` from `gasPrice`. Not sure if this is right? @ethanfrey

src/binary/ibc-setup/commands/balances.ts
bartmacbartek: https://github.com/confio/ts-relayer/pull/91/files#diff-58673c679ff81312a0f7389a4cf697deac398637a352094bd9d4832a5cf28ee4R49

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,79 @@
...
+        const coin = await client.query.bank.unverified.balance(
+          address,
+          gasPrice.denom
+        );
+
+        return [chain, coin];
+      })
+    )
+  )
+    .filter(([, coin]) => Number(coin.amount) > 0)
```
bartmacbartek, line 74: I was hoping to get something form `@cosmjs/math` but nothing really worked. I thought of using `Decimal` but I would have to parse the amount in some way to provide fractional digits.

Ended up using `Number` since the precision is not required here, right?

src/binary/ibc-setup/commands/balances.ts
bartmacbartek: I think something went wrong here. I have the parsing already with `GasPrice`: https://github.com/confio/ts-relayer/pull/91/files#diff-58673c679ff81312a0f7389a4cf697deac398637a352094bd9d4832a5cf28ee4R53-R67

Does it look okay?

src/binary/ibc-setup/commands/balances.ts
ethanfrey: We can just return this unnormalized, eg `1234000 uatom` (what we get from the response) rather than `1.234 ATOM` (what the wallet displays).

src/binary/ibc-setup/commands/balances.ts
ethanfrey: Oh, yeah don't use that one.
We should have the same coin in `../../../codec/cosmos/...`

But that's a proto type with no special functionality. If you need the interface for types, use the local one

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,79 @@
...
+export async function run(options: Options) {
+  const registryFilePath = path.join(options.home, registryFile);
+  const registry = loadAndValidateRegistry(registryFilePath);
+
+  const addresses = await getAddresses(registry.chains, options.mnemonic);
+
+  const balances = (
+    await Promise.all(
+      addresses.map<Promise<[string, Coin]>>(async ([chain, data, address]) => {
+        const signer = await DirectSecp256k1HdWallet.fromMnemonic(
```
ethanfrey, line 47: These lines (47-63) make a IbcClient from the chain data. This is a great function to pull out.

You don't need to even provide address here, you can get that from the signer (like in the getAddresses function)

src/binary/ibc-setup/commands/balances.ts
ethanfrey: perfect

src/binary/ibc-setup/commands/balances.ts
ethanfrey: If you do that, you could then do something like:

```ts
registry.chains.map(([chain, data]) => {
  const {client, address, feeToken} = await buildIbcClient(data);
  const coin = await client.query.bank.unverified.balance(address, feeToken);
  return [chain, coin];
})....
```

ethanfrey: But yeah, that can also be in another PR (to pull it out).

Given one chain data to build the IbcClient will be needed for most of the other functions. You can merge and leave it to Will to refactor this for his command

src/binary/ibc-setup/commands/balances.ts
ethanfrey: You could also simply check if `coin.amount !== "0"`

It can never be negative

ethanfrey (commented):
Looking good. 
Could simplify the logic a bit, but let's pull it out of draft

ethanfrey (approved):

src/binary/ibc-setup/commands/balances.ts
bartmacbartek: The thing is if we want to display only non-zero balances (as in the spec) I need to filter it through `coin.amount > 0`. However, the `coin.amount` is a string and I cannot compare it with numbers so I'm forced to convert it to one.

ethanfrey: It works like this

bartmacbartek: Yeah, I know this for precision :)

Changed to `!== "0"`. Smart! 4b61b97

src/binary/ibc-setup/commands/balances.ts
bartmacbartek: f9def0b

src/binary/ibc-setup/commands/balances.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+function fakeBalance(amount: string) {
+  return (Promise.resolve({
+    query: {
+      bank: {
+        unverified: {
+          balance: sinon.fake.returns({ amount, denom: 'sampledenom' }),
+        },
+      },
+    },
+  }) as unknown) as Promise<IbcClient>;
```
bartmacbartek, line 26: Casting is required to fulfill the huge `IbcClient` interface.

willclarktech: 👍 Do whatever you need in test files.

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,86 @@
...
+import { registryFile } from '../../constants';
+import { loadAndValidateApp } from '../utils/load-and-validate-app';
+import { loadAndValidateRegistry } from '../utils/load-and-validate-registry';
+import { resolveHomeOption } from '../utils/options/shared/resolve-home-option';
+import { resolveKeyFileOption } from '../utils/options/shared/resolve-key-file-option';
+import { resolveMnemonicOption } from '../utils/options/shared/resolve-mnemonic-option';
+
+import { Flags, getAddresses, Options } from './keys-list';
+
+export { Options }; // re-export for testing
```
willclarktech, line 19: Why not just import it from the same place in tests?

src/binary/ibc-setup/commands/balances.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+
+import { IbcClient } from '../../../lib/ibcclient';
+import { generateMnemonic } from '../utils/generate-mnemonic';
+
+import { Options, run } from './balances';
+
+const consoleLog = sinon.stub(console, 'log');
+const fsReadFileSync = sinon.stub(fs, 'readFileSync');
+const ibcClient = sinon.stub(IbcClient, 'connectWithSigner');
+const mnemonic = generateMnemonic();
```
willclarktech, line 15: Probably best to hardcode one here for reproducibility/transparency.

src/binary/ibc-setup/commands/balances.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function fakeBalance(amount: string) {
```
willclarktech, line 17: Naming: `createFakeIbcClient`? Also if you make this an `async` function I don't think you need the `Promise.resolve` and the cast can be simpler.

src/binary/ibc-setup/commands/balances.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    .onCall(0)
+    .returns(fakeBalance('1'))
+    .onCall(1)
+    .returns(fakeBalance('2'))
+    .onCall(2)
+    .returns(fakeBalance('3'));
+
+  await run(options);
+
+  t.assert(fsReadFileSync);
```
willclarktech, line 81: ?

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,86 @@
...
+          address,
+          gasPrice.denom
+        );
+
+        return [chain, coin];
+      })
+    )
+  )
+    .filter(([, coin]) => coin.amount !== '0')
+    .map(([chain, coin]) => `${chain}: ${coin.amount}`)
```
willclarktech, line 77: We need to print the `denom` here as well. Eg 
```ts
`${chain}: ${coin.amount}${coin.denom}`
```

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -0,0 +1,86 @@
...
+    .join(os.EOL);
+
+  if (!balances) {
+    console.log('No funds found on any balance.');
```
willclarktech, line 81: Suggestion: "No funds found for default denomination on any chain."

src/binary/ibc-setup/commands/balances.ts
bartmacbartek: 1ad4c2d

src/binary/ibc-setup/commands/balances.ts
bartmacbartek: 6b83962

src/binary/ibc-setup/commands/balances.spec.ts
bartmacbartek: 😱 7688a11

src/binary/ibc-setup/commands/balances.spec.ts
bartmacbartek: done in 3ce03b0

src/binary/ibc-setup/commands/balances.spec.ts
bartmacbartek: done in 3ce03b0

src/binary/ibc-setup/commands/balances.ts
bartmacbartek: 765bc7f

bartmacbartek: I thought it's cleaner to import `Options` from `balances.ts` instead of `keys-list.ts`  in `balances.spec.ts`.


But it doesn't really matter, changed.

src/binary/ibc-setup/commands/balances.spec.ts
```diff
@@ -15,7 +15,7 @@ const mnemonic =
   'accident harvest weasel surge source return tag supreme sorry isolate wave mammal';
 
 function buildIbcArgs(rpc: string) {
-  return [rpc, { getAccounts: {}, signDirect: {} }, '', {}] as const;
+  return [rpc, sinon.match.any, sinon.match.any, sinon.match.any] as const;
```
willclarktech, line 18: 👌

PR #92, opened 2021-03-02
Use yarn.lock for cache checksum

bartmacbartek:
- use `yarn.lock` for the checksum cache
- replace `install yarn` step with `set yarn version to latest stable` since `yarn` comes preinstalled on the machine we're using (speeds up the running time by 11 seconds 🚀 😅)


willclarktech (approved):
Nice, this is what we do in CosmJS.

PR #93, opened 2021-03-03
Update chain IDs and regenerate simapp template

willclarktech:


ethanfrey:
Why do this?

Now both have revision 1? All my revision number tests were based on the fact that simapp and wasmd had different revision numbers. It is good we have different cases on both sides.

PR #94, opened 2021-03-04
56 ibc setup ics20

bartmacbartek:


src/binary/ibc-setup/commands/ics20.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+const consoleLog = sinon.stub(console, 'log');
+
+const mnemonic =
+  'enlist hip relief stomach skate base shallow young switch frequent cry park';
+
+const registryYaml = `
+version: 1
+
+chains:
+  musselnet:
```
ethanfrey, line 23: You can remove musselnet from all these tests, we will only use the two local (ci) chains

src/binary/ibc-setup/commands/ics20.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+`
+  );
+  t.assert(fsWriteFileSync.calledOnce);
+  t.is(args[0], path.join(options.home, appFile));
+  t.regex(args[1], contentsRegexp);
+  t.assert(consoleLog.calledTwice);
+  t.assert(consoleLog.calledWithMatch(/Created connections/));
+  t.assert(consoleLog.calledWithMatch(/Created channels/));
+
+  const nextAllConnectionsWasm = await ibcClientWasm.query.ibc.connection.allConnections();
```
ethanfrey, line 106: Nice tests

src/binary/ibc-setup/commands/ics20.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+};
+
+test.beforeEach(() => {
+  sinon.reset();
+});
+
+test.serial('ics20 create channels with new connection', async (t) => {
+  const ibcClientWasm = await createClient(mnemonic, {
+    prefix: 'wasm',
+    // hd_path: `m/44'/108'/0'/2'`,
```
ethanfrey, line 62: hd path is totally optional for these tests. nice to have to ensure addresses are different on both sides, but it doesn't really matter what key we use there

(You can just make a follow up issue later if this is a pain)

src/binary/ibc-setup/commands/ics20.ts
```diff
@@ -0,0 +1,152 @@
...
+  if (!destChain) {
+    throw new Error('dest chain not found in registry');
+  }
+  const ordering = Order.ORDER_UNORDERED;
+  const version = 'ics20-1';
+
+  const nodeA = await createClient(options.mnemonic, srcChain);
+  const nodeB = await createClient(options.mnemonic, destChain);
+  // TODO: Handle if connection flag is provided
+  const link = await Link.createWithNewConnections(nodeA, nodeB);
```
ethanfrey, line 102: Nice

src/binary/ibc-setup/commands/ics20.ts
```diff
@@ -0,0 +1,152 @@
...
+  console.log(
+    `Created connections ${link.endA.connectionID} (${link.endA.clientID}) <=> ${link.endB.connectionID} (${link.endB.clientID})`
+  );
+
+  const srcClient = link.endA.clientID;
+  const destClient = link.endB.clientID;
+  const srcConnection = link.endA.connectionID;
+  const destConnection = link.endB.connectionID;
+  const appFilePath = path.join(options.home, appFile);
+  const appYaml = yaml.dump(
```
ethanfrey, line 112: I like how it just appends to what was there

src/binary/ibc-setup/commands/ics20.ts
```diff
@@ -0,0 +1,152 @@
...
+      destConnection,
+    },
+    {
+      lineWidth: 1000,
+    }
+  );
+
+  fs.writeFileSync(appFilePath, appYaml, { encoding: 'utf-8' });
+
+  const channels = await link.createChannel(
```
ethanfrey, line 127: :+1:

ethanfrey (commented):
Nice stuff

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -46,7 +46,7 @@ export async function run(options: Options) {
       addresses.map<Promise<[string, Coin]>>(async ([chain, data, address]) => {
         const signer = await DirectSecp256k1HdWallet.fromMnemonic(
           options.mnemonic,
-          stringToPath(data.hd_path),
+          data.hd_path ? stringToPath(data.hd_path) : undefined,
```
ethanfrey, line 49: nice optional

src/binary/ibc-setup/commands/ics20.ts
```diff
@@ -68,6 +104,18 @@ export async function ics20(flags: Flags): Promise<void> {
...
+    app.srcConnection,
+    process.env.RELAYER_SRC_CONNECTION
+  );
+  const destConnection = resolveOption(
+    flags.destConnection,
+    app.destConnection,
+    process.env.RELAYER_DEST_CONNECTION
+  );
+
+  const connections = resolveConnections(srcConnection, destConnection);
```
ethanfrey, line 118: Straight forward usage here to combine them

src/binary/ibc-setup/commands/ics20.ts
```diff
@@ -77,11 +125,37 @@ export async function ics20(flags: Flags): Promise<void> {
       mnemonic,
       srcPort,
       destPort,
+      connections,
     },
     app
   );
 }
 
+async function resolveLink(
```
ethanfrey, line 134: :+1:

src/binary/ibc-setup/commands/ics20.spec.ts
```diff
@@ -107,32 +90,38 @@ destConnection: .+
...
   const srcConnectionId = srcConnectionIdMatch?.groups?.connection;
   assert(srcConnectionId);
-  const srcClientMatch = /srcClient: (?<srcClient>.+)/.exec(args[1]);
-  const srcClient = srcClientMatch?.groups?.srcClient;
-  const nextConnectionWasm = await ibcClientWasm.query.ibc.connection.connection(
-    srcConnectionId
-  );
+  // const srcClientMatch = /srcClient: (?<srcClient>.+)/.exec(args[1]);
+  // const srcClient = srcClientMatch?.groups?.srcClient;
+  // const nextConnectionWasm = await ibcClientWasm.query.ibc.connection.connection(
```
ethanfrey, line 95: I see why you removed the client match (might as well just delete it not comment it out).

But why is `nextConnectionWasm` commented out? It should still work, right?

ethanfrey (approved):
I'd just enable one tests you commented out. Otherwise, let's pull out of draft and merge it

ethanfrey:
Ah, there is a real failing test, not just the header time issue

src/binary/ibc-setup/commands/ics20.spec.ts
bartmacbartek: 4dfc837

src/binary/ibc-setup/commands/ics20.spec.ts
bartmacbartek: I made `hd_path` optional in the `registry.yaml` so we can use defaults in tests.

Also: https://github.com/confio/ts-relayer/pull/94/files/1269e7e9a3a0090d7431e0d115e947af07b59aa4..d53fb34f8cb1d82a7fea5e9254b7af39c7521fb1#diff-58673c679ff81312a0f7389a4cf697deac398637a352094bd9d4832a5cf28ee4R49

src/binary/ibc-setup/commands/ics20.ts
bartmacbartek: Not sure about one case:

**app.yaml**:
```
...
srcConnection: 'some-src-connection-1'
destConnection: 'some-dest-connection-1'
```

running `ibc-setup ics20 --srcConnection test-conn-1 --destConnection test-conn-1 ...`

will result in overriding `some-src-connection-1` and `some-dest-connection-1`. Is that the desired behavior?

src/binary/ibc-setup/commands/ics20.spec.ts
bartmacbartek: I kept `nextConnection*` checks: https://github.com/confio/ts-relayer/pull/94/files#diff-5bb34d24bff296d595d81dafe53eb04885ccbf17a096785882bdc8be02806a17R116-R117

but only as simple asserts since I cannot capture `clientId` from the test anymore.

src/binary/ibc-setup/commands/ics20.ts
ethanfrey: All channels must be on the same connections.
If we change it, the old ones will be invalid. If we ignore them, then the new channel cannot be made.

I bet this argument isn't good.

For the "advanced" manual channel creation commands, maybe we can allow input from app.yaml, flags, or env (assuming they don't edit app.yaml automatically). However, for this use case (ics20 should be easy to use), let us *only* take src/destConnection from `app.yaml`

----

TL;DR: just change the following


```ts
  const srcConnection = resolveOption(
    flags.srcConnection,
    process.env.RELAYER_SRC_CONNECTION,
    app.srcConnection
  );
  const destConnection = resolveOption(
    flags.destConnection,
    process.env.RELAYER_DEST_CONNECTION,
    app.destConnection
  );
```

to:

```ts
  const srcConnection = app.srcConnection;
  const destConnection = app.destConnection;
```

And unregister the flags and update the spec doc (maybe with an explanation). Then ready to merge

dontmakemethink: 48a9748
577b229

PR #95, opened 2021-03-08
53 connection setup

### LINKED ISSUE #53: Add connection setup
Builds on #52 

* Add [connection setup](https://github.com/confio/ts-relayer/blob/main/spec/ibc-setup.md#connection-setup) to `ibc-setup`
  * Code just can call `Link.createWithNewConnections()` with proper `IbcClients`
* Add simple test for CI

bartmacbartek:
Closes #53

### DIFF
--- a/src/binary/ibc-setup/commands/connect.ts
+++ b/src/binary/ibc-setup/commands/connect.ts
@@ -1,0 +1,69 @@
+import path from 'path';
+
+import { registryFile } from '../../constants';
+import { loadAndValidateApp } from '../utils/load-and-validate-app';
+import { loadAndValidateRegistry } from '../utils/load-and-validate-registry';
+import { resolveRequiredOption } from '../utils/options/resolve-required-option';
+import { resolveHomeOption } from '../utils/options/shared/resolve-home-option';
+import { resolveKeyFileOption } from '../utils/options/shared/resolve-key-file-option';
+import { resolveMnemonicOption } from '../utils/options/shared/resolve-mnemonic-option';
+
+export type Flags = {
+  readonly interactive: boolean;
+  readonly home?: string;
+  readonly keyFile?: string;
+  readonly mnemonic?: string;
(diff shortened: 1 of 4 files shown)


src/binary/ibc-setup/commands/connect.ts
```diff
@@ -0,0 +1,103 @@
...
+      destConnection: link.endB.connectionID,
+    },
+    {
+      lineWidth: 1000,
+    }
+  );
+
+  fs.writeFileSync(path.join(options.home, appFile), appYaml, {
+    encoding: 'utf-8',
+  });
```
bartmacbartek, line 98: @ethanfrey 

Let's say I use
```
ibc-setup connect --src local_wasm --dest local_simapp --mnemonic xxxx
```

while my **app.yaml** looks like this
```
src: some_chain
dest: another_chain
mnemonic: enlist hip relief stomach skate base shallow young switch frequent cry park
```

Should the `src`, `dest`, `mnemonic` properties be overridden? Adding `srcConnection` and `destConnection` next to the not-related `src` and `dest` chains doesn't make sense, right?

---

Maybe we should write only when **app.yaml** doesn't exist?

---

src/binary/ibc-setup/commands/connect.ts
```diff
@@ -0,0 +1,103 @@
...
+  readonly src: string;
+  readonly dest: string;
+};
+
+export async function connect(flags: Flags) {
+  const home = resolveHomeOption({ homeFlag: flags.home });
+  const app = loadAndValidateApp(home);
+  if (!app) {
+    throw new Error(`${appFile} not found at ${home}`);
+  }
```
bartmacbartek, line 39: This is probably not right, related: https://github.com/confio/ts-relayer/pull/95/files#r590706164

ethanfrey (approved):
Looks good and very valid points.

Let's discuss this more in #98 but merging as is.

Issue #54, opened 2021-02-18
Add Channel setup functionality


ethanfrey:
Builds on #53 

The following just need the registry (should work if mnemonic is not provided):

* Implement `ibc-setup ports --chain=ABC`
* Implement `ibc-setup connections --chain=ABC`
* Implement `ibc-setup channels --chain=ABC [--port=transfer]`

The following needs the mnemonic as well, and will reuse an existing connection via `Link.createWithExistingConnections`

* Implement `ibc-setup channel --src=ABC --dest=XYZ --src_connection=connection-3 --dest_connection=connection-4 ...`

(All commands [defined in spec](https://github.com/confio/ts-relayer/blob/main/spec/ibc-setup.md#channel-setup))

Please add a test: create connection as in #53, query channels with nothing, make new channel on existing connection, query channels that it is there

bartmacbartek:
`ibc-setup channels --chain=ABC [--port=transfer]` (point 3) is merged via https://github.com/confio/ts-relayer/pull/97 but it still lacks tests

PR #97, opened 2021-03-09
Add ibc-setup channels

bartmacbartek:
addresses #54


src/binary/ibc-setup/commands/channels.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    return;
+  }
+
+  const output = [
+    ['CHANNEL_ID', 'PORT', 'STATE'].map(appendSpaces).join(''),
+    ...channels,
+  ].join(os.EOL);
+
+  console.log(output);
```
bartmacbartek, line 146: @ethanfrey 

The spec says:
>It shows all channels, both open, as well as channels in the handshaking process or closed ones (with a comment on non-open ones)

but here we turn every state into text, like:
image

Is it okay?

src/binary/ibc-setup/commands/channels.ts
```diff
@@ -0,0 +1,147 @@
...
+  await run(options);
+}
+
+function assureMnemonic(mnemonic: string | null) {
+  if (!mnemonic) {
+    return generateMnemonic();
+  }
+
+  return mnemonic;
+}
```
bartmacbartek, line 69: Extracted this into a separate function to not hit `generateMnemonic()` until it's necessary.

src/binary/ibc-setup/commands/channels.ts
```diff
@@ -0,0 +1,147 @@
...
+    case State.UNRECOGNIZED:
+    default:
+      return 'Unrecognized';
+  }
+}
+
+function appendSpaces(value: string) {
+  const spaces = new Array(24 - value.length).fill(' ').join('');
+  return `${value}${spaces}`;
+}
```
bartmacbartek, line 97: This one is temporary, we'll get rid of it quickly after putting the logger in place.

src/binary/ibc-setup/commands/channels.ts
ethanfrey: That output looks quite nice

src/binary/ibc-setup/utils/options/shared/resolve-mnemonic-option.ts
```diff
@@ -46,17 +47,30 @@ type Params = {
...
-}: Params) {
-  return resolveRequiredOption('mnemonic')(
+export async function resolveMnemonicOption(
+  params: Params,
+  optional: true
+): Promise<string | null>;
+export async function resolveMnemonicOption(
+  params: Params,
+  optional?: false
+): Promise<string>;
```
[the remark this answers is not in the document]
ethanfrey: Interesting

src/binary/ibc-setup/commands/channels.ts
```diff
@@ -0,0 +1,147 @@
...
+  const registry = loadAndValidateRegistry(registryFilePath);
+
+  const chain = registry.chains[options.chain];
+  if (!chain) {
+    throw new Error(`Chain ${options.chain} not found in ${registryFilePath}.`);
+  }
+
+  const mnemonic = assureMnemonic(options.mnemonic);
+
+  const client = await createClient(mnemonic, {
```
ethanfrey, line 110: Yeah, we could make some read-only client class so you don't need a fake mnemonic here.
But that would be some work. Nice to just generate a fake one to make it all run.

src/binary/ibc-setup/commands/channels.ts
```diff
@@ -0,0 +1,147 @@
...
+    hd_path: chain.hd_path,
+  });
+
+  const {
+    channels: allChannels,
+  } = await client.query.ibc.channel.allChannels();
+
+  const channels = allChannels
+    .filter(
+      (channel) => (options.port ? channel.portId === options.port : true) // don't filter if port is not specified
```
ethanfrey, line 122: I ended up doing something in the SDK recently for internal queries. Even with access to all the Go code, there was no way to get all channels by port without iterating over all of them,

ethanfrey (commented):
First glance looks good.

Since I really want to get some stuff running, if you have manually tested this, feel free to merge, but **do not** close the original issue. Just say "part 1 of 54" or something and then add a comment on the issue that this needs test coverage.

Let's get on with the logging, I am sure I will need this to help debug soon.

ethanfrey (approved):

Issue #89, opened 2021-03-02
Improve test suite


willclarktech:
This issue is for gathering potential improvements for the test suite.

- [ ] Run tests in parallel where possible. (ava is very good for this.)
- [ ] Move test suite out of `testutils.spec.ts`

ethanfrey:
> Move test suite out of testutils.spec.ts

I would simply remove it. These helpers are stable and now covered indirectly from everything else. It was there to bootstrap

ethanfrey:
> Run tests in parallel where possible. 

Yup, Most tests that don't do things like "queryAllClients()" should be runable in parallel. We can create 2 different connections with different ids at the same time.

It will also be a good test of the relayer in more difficult, real world situations.

What will likely break:
* Funding wallets (We submit multiple bank.send tx from the same account. It cannot sign more that one tx per block. We just need to make this controlled by a Mutex or something)
* Tests that use `allXyz()` queries - state changes outside of their control

I assume all tests make their own clients, so all connection/channel/packet tests are sandboxed. We could later test two relayers running in parallel on the same connection to ensure it is solid (different issue)

PR #99, opened 2021-03-10
Relayer Loop

### LINKED ISSUE #12: Create relay loop
- **I'm submitting a ...**
  [ ] bug report
  [x] feature request
  [ ] question about the decisions made in the repository
  [ ] question about how to use this project

- **Summary**

Builds on #9 and #10

See [Loop from spec](https://github.com/confio/ts-relayer/blob/main/spec/ibc-relayer.md#loop), and implement with polling:

* Sleep N seconds
* Query the current block height/time
    * If too far from last client update -> Update Client
* Query src, dest for all send_packet events since H1 and add to Pending Packets
  * Set H1 to new height
  * Relay on opposite chain
* Query src, dest for all acknowledge_packet events since H2 and add to Pending Packets
  * Set H2 to new height
  * Relay on opposite chain

ethanfrey:
Most of #12 
Part 1 of #89 

- [x] Implement and test the base logic used in each step of the loop inside Link.
- [x] Write and test a single-short relayer (add `--once` flag)

Follow up PR:
- [ ] Have infinite loop in relayer


src/binary/ibc-relayer/commands/start.ts
```diff
@@ -81,12 +92,17 @@ export async function start(flags: Flags) {
...
     destConnection,
+    // TODO: make configurable
+    runOnce: true,
+    // once per day: 86400s
+    maxAgeA: 86400,
+    maxAgeB: 86400,
   };
 
-  run(options, logger);
+  await run(options, logger);
```
bartmacbartek, line 102: no need for `await` here but it doesn't hurt

ethanfrey: I guess not, but if I neither `await` nor return the Promise, then if the caller tries `await`, it will not wait for the final promise with the heart of the work.

I guess we don't do it that way here, but it is a habit after a few gotchas with unawaited promises

src/lib/link.spec.ts
```diff
@@ -1,18 +1,18 @@
...
 import {
   ics20,
-  randomAddress,
   setup,
   simapp,
   TestLogger,
+  transferTokens,
   wasmd,
-} from './testutils.spec';
+} from './testutils';
```
willclarktech, line 15: @ethanfrey This filename had a `.spec` in it to group it with test files, eg so that it's filtered out of the files published to npm.

ethanfrey: When I removed the unit tests there and left it as spec, ava complained (errored).

Maybe just add this file to the npmignore file?

PR #101, opened 2021-03-10
58 logger for ibc-relayer start

bartmacbartek:


src/binary/ibc-relayer/commands/start.ts
```diff
@@ -29,6 +34,14 @@ type Options = {
 };
 
 export async function start(flags: Flags) {
+  const logLevel = resolveOption(
+    flags.logLevel,
+    process.env.RELAYER_LOG_LEVEL,
+    'info'
```
ethanfrey, line 40: :+1:

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -29,6 +34,14 @@ type Options = {
...
+  );
+
+  const logger = createLogger(logLevel as Level);
```
ethanfrey, line 43: I would do some validation here not naively casting string to a set of valid strings.
But yeah, for the next PR

src/binary/ibc-relayer/index.ts
```diff
@@ -28,6 +28,19 @@ program
   .addOption(mnemonicOption)
   .option('--src-connection <connection>')
   .option('--dest-connection <connection>')
+
+  .addOption(
+    new Option('--log-level <level>').choices([
```
ethanfrey, line 33: Ah, never mind, it looks like you do enforce this. Nice

src/binary/types.ts
```diff
@@ -19,3 +19,9 @@ export type AppConfig = {
   mnemonic?: string;
   keyFile?: string;
 };
+
+export type LoggerFlags = {
+  logLevel?: string;
+  verbose: boolean;
```
ethanfrey, line 25: For the future right?
Not needed for my debugging now.

ethanfrey (approved):
Thanks

PR #104, opened 2021-03-10
Finish relayer loop

### LINKED ISSUE #12: Create relay loop

ethanfrey:
Closes #12 

Follow up from #99 

- [x] Have infinite loop in relayer
- [x] Allow timing configuration via flags (or config)
- [x] Better configuration of trusted period when creating clients (should be ~ 2/3 of unbonding period, 2 weeks)

Some pending optimizations pushed to #110

### DIFF
--- a/src/binary/ibc-relayer/commands/start.ts
+++ b/src/binary/ibc-relayer/commands/start.ts
@@ -128,8 +128,6 @@ export async function start(flags: Flags) {
     once,
   };
 
-  logger.verbose('Starting relayer with options', options);
-
   await run(options, logger);
 }
 
--- a/src/binary/ibc-relayer/commands/start.ts
+++ b/src/binary/ibc-relayer/commands/start.ts
@@ -28,7 +28,8 @@ type Flags = {
 
 // TODO: do we want to make this a flag?
 type LoopOptions = {
-  runOnce: boolean;
+  // how many seconds we sleep between relaying batches
+  pollingFrequency: number;
   // number of seconds old the client on chain A can be
   maxAgeA: number;
   // number of seconds old the client on chain B can be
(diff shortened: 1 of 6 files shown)


src/binary/ibc-relayer/commands/start.ts
```diff
@@ -24,15 +26,18 @@ type Flags = {
   mnemonic?: string;
   srcConnection?: string;
   destConnection?: string;
+  poll?: number;
+  maxAgeSrc?: number;
+  maxAgeDest?: number;
 } & LoggerFlags;
```
bartmacbartek, line 32: `Partial<LoopOptions>` could be used here instead.

I'm going to clean this up anyway so we can merge.

bartmacbartek (approved):

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -17,6 +17,17 @@ import { resolveKeyFileOption } from '../../utils/options/shared/resolve-key-fil
...
 
+type LoopOptions = {
+  // how many seconds we sleep between relaying batches
+  poll: number;
+  // number of seconds old the client on chain A can be
+  maxAgeSrc: number;
+  // number of seconds old the client on chain B can be
+  maxAgeDest: number;
+  // if set to 'true' quit after one pass
+  once?: boolean;
```
bartmacbartek, line 28: The question mark `?` is redundant. Parser always resolves non-parametrized flags to booleans.

PR #106, opened 2021-03-11
58-logger

### LINKED ISSUE #58: Set log-level and basic logger in binary
Requires #57 

It should implement [the logger spec](https://github.com/confio/ts-relayer/blob/main/spec/logging.md). The only needed transports are to console/stdout and to file. Assuming a human-readable format on console, JSON if written to file (so we can send it to Kibana, etc)

This should be enough for the first release, we can revisit more transports and format options in v2 based on user feedback

bartmacbartek:
Closes #58


src/binary/create-logger.ts
```diff
@@ -1,16 +1,52 @@
...
 };
+type Level = keyof typeof levels;
+
+export const defaultLevel = 'info'; // if not provided
+
+function validateLevel(level: string | null): level is Level {
+  return level ? Object.keys(levels).includes(level) : false;
+}
+
+export function resolveLevel(
```
ethanfrey, line 21: It took me a few passes, but this does seem like solid logic and doing what was specified

ethanfrey (commented):
Nice work parsing the levels.

src/binary/create-logger.ts
bartmacbartek: I can't read this either but it's well tested --> https://github.com/confio/ts-relayer/pull/106/files#diff-b3d42f480f31b66005903f1d5e9e38cebbf3df590a009b34edadd0acb793da81

Maybe we should simplify the spec to throw if multiple levels are provided, e.g.:
```bash
ibc-setup start --log-level verbose -qv # throw 
```

It would make the parsing easier.

ethanfrey: It looks good as it is. And the tests are top

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -76,9 +77,9 @@ export async function run(options: Options) {
     .join(os.EOL);
 
   if (!balances) {
-    console.log('No funds found for default denomination on any chain.');
+    logger.info('No funds found for default denomination on any chain.');
```
ethanfrey, line 80: I would actually use consol.log on these pieces. The stuff that is supposed to go to stdout for interactive commands.  All the details of running should go to logger

ethanfrey (approved):
Looks good. Minor commend, then please merge

ethanfrey:
This is already a huge help. I'm going to merge it, and leave the console.log item for later.

Ideally I can run `./ibc-setup ics20 -q` and only get the last notifications (the channel created - never filtered as it is not with the logger, it is the intended output

PR #113, opened 2021-03-12
Add relayer test networks to demo/registry.yaml

orkunkl:
This PR adds the network details that we launched for testing ts-relayer on multi node networks.


ethanfrey (approved):
thanks

PR #117, opened 2021-03-13
Unit tests to ensure timeout calculated properly

ethanfrey:
There were some issues, let's track them down.
So far, the current code looks proper.

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -101,7 +101,7 @@ export function timestampFromDateNanos(
 export function secondsFromDateNanos(
   date: ReadonlyDateWithNanoseconds
 ): number {
-  return date.getTime() / 1000;
+  return Math.floor(date.getTime() / 1000);
 }
 
 export function buildConsensusState(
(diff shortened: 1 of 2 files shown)


src/lib/utils.ts
```diff
@@ -315,7 +314,7 @@ export function multiplyCoin({ amount, denom }: Coin, mult: number): Coin {
 }
 
 // return true if a > b, or a undefined
-function heightGreater(a: Height | undefined, b: Height): boolean {
+export function heightGreater(a: Height | undefined, b: Height): boolean {
```
bartmacbartek, line 317:
```suggestion
export function heightGreater(a?: Height, b: Height): boolean {
```

src/lib/utils.ts
```diff
@@ -333,7 +332,7 @@ function heightGreater(a: Height | undefined, b: Height): boolean {
 
 // return true if a > b, or a 0
 // note a is nanoseconds, while b is seconds
-function timeGreater(a: Long | undefined, b: number): boolean {
+export function timeGreater(a: Long | undefined, b: number): boolean {
```
bartmacbartek, line 335:
```suggestion
export function timeGreater(a?: Long, b: number): boolean {
```

bartmacbartek (approved):
Looks good to me, approved with syntax suggestions.

src/lib/utils.ts
ethanfrey: Nope. You cannot have a required argument after an optional argument - tsc already caught me on this one (I tried it)

ethanfrey:
Since `tsc` and `eslint` require the syntax I used, I will keep it (I do prefer your suggestions)

PR #118, opened 2021-03-14
Log stacktrace in relayer loop

alpe:
* Log full stacktrace with error message


ethanfrey (approved):
This is much better. Thanks.

JS tip: backquotes ` are only needed if you use variable interpolation inside the string (`${e}`). Otherwise you can use normal single quotes, '

No problem, it is just like doing `fmt.Sprintf("Caught error: ")`

PR #119, opened 2021-03-15
Refactor options resolvers

bartmacbartek:
- refactor resolvers

  Removed
  - `resolveRequiredOption.ts`
  - `resolveNumericOption.ts`
  - `resolveRequiredNumericOption.ts`
  
  in favor of more functional (but tested) `resolveOption.ts`.
  
- centralize validation and parsing
- add custom exceptions for better error handling in the future


src/binary/ibc-relayer/index.ts
```diff
@@ -15,14 +15,6 @@ import { loggerWithErrorBoundary } from '../utils/logger-with-error-boundary';
 
 import { start } from './commands/start';
 
-function parseNumber(value: string) {
-  const parsed = parseInt(value, 10);
-  if (isNaN(parsed)) {
-    throw new InvalidOptionArgumentError('Not a number.');
-  }
-  return parsed;
-}
```
bartmacbartek, line 24: Dropping the validation at the commander level in favor of the `resolveOption` function.

e.g.:
```ts
const max = resolveOption('max', { required: true, integer: true })(flags.max, app?.max, process.env.RELAYER_MAX)
// "max" is of type number
```

Otherwise, we would have to maintain the parsing & validation in 2 places because the value can come from different sources such as app config or the environment variable (string).

Also, having it centralized gives us control over the error messages.

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -66,42 +73,43 @@ export async function start(flags: Flags, logger: Logger) {
...
-  const destConnection = resolveRequiredOption('destConnection')(
+  const destConnection = resolveOption('destConnection', { required: true })(
     flags.destConnection,
     app?.destConnection,
     process.env.RELAYER_DEST_CONNECTION
   );
 
   // TODO: add this in app.yaml, process.env
-  const poll = resolveRequiredNumericOption('poll')(
+  const poll = resolveOption('poll', { required: true, integer: true })(
```
ethanfrey, line 100: Ah, nice use of second arg of options

src/binary/utils/logger-with-error-boundary.ts
```diff
@@ -12,6 +12,14 @@ export function loggerWithErrorBoundary<T>(
     try {
       await command(flags, logger);
     } catch (error) {
+      /*
```
ethanfrey, line 15: Nice to add this. If you include the argument name when creating InvalidOptionError, then maybe it makes sense to do some special formatting here. If you create a nice message, I don't see the need

src/binary/utils/options/resolve-option.ts
```diff
@@ -1,6 +1,67 @@
+import { InvalidOptionError } from '../../exceptions/InvalidOptionError';
+
+type Args<T = string> = Array<(T | undefined | null) | (() => T | null)>;
+
+// if required and integer then it's a number
+export function resolveOption(
```
ethanfrey, line 6: Nice use of type overloads.

ethanfrey (approved):
Clean.

src/binary/utils/logger-with-error-boundary.ts
bartmacbartek: I'll leave the comment for now and remove it in a future iteration.

PR #120, opened 2021-03-15
Parallel test runs

### LINKED ISSUE #89: Improve test suite

ethanfrey:
Closes #89 

This works faster but has 2 issues:

* Since test cases run in separate processes, the mutex only prevents conflicts *within one file*. We need to see how to prevent conflicts between files
* Sinon doesn't work to well in these parallel situations. (remove .serial from the binary tests to see this)


willclarktech:
> Sinon doesn't work to well in these parallel situations.

Without looking into this my guess is we can solve whatever the problem is with sandboxes: https://sinonjs.org/releases/v9.2.4/sandbox/

webmaster128:
Looking at #254, can we close this?

ethanfrey:
Let's close the PR for sure.

Maybe revisit the issue with the new framework (which must run everything sequentially at first until we refactor the startup code)

PR #121, opened 2021-03-15
110 store last heights

### LINKED ISSUE #110: Persist last queried heights over relayer restart
Follow up from #104

* Relayer persistently stores last height queried for better performance on startup

Relevant lines of code from `ibc-relayer`:

```ts
async function relayerLoop(link: Link, options: LoopOptions, logger: Logger) {
  // TODO: fill this in with real data on init
  // (how far back do we start querying... where do we store state?)
  let nextRelay = {};
```

On first start, it can query all packets from genesis, which can be slow. It should then persist this state in a file (not app.yaml, but in the same dir). This file is read on restart and is used, so it just queries since the last time it ran.

The file is not meant to be human-readable (just delete it if you want to force scanning all events since forever), and could be stored in JSON rather than YAML if that is easier. 

Also, it would be nice to allow a user to set those values initially even if there is no JSON file. These take presidence over any json file:
* `--scan-from-src <height>` - equivalent to setting `packetHeightA` and `ackHeightA`
* `--scan-from-dest <height>` - equivalent to setting `packetHeightB` and `ackHeightB`

hmm... maybe we should rename those variables in the JSON?

bartmacbartek:
Closes #110

### DIFF
--- a/src/binary/ibc-relayer/commands/start.ts
+++ b/src/binary/ibc-relayer/commands/start.ts
@@ -17,16 +17,14 @@ import { resolveKeyFileOption } from '../../utils/options/shared/resolve-key-fil
 import { resolveMnemonicOption } from '../../utils/options/shared/resolve-mnemonic-option';
 import { signingClient } from '../../utils/signing-client';
 
+type ResolveHeightsParams = {
+  scanFromSrc: number | null;
+  scanFromDest: number | null;
+  home: string;
+};
+
 function resolveHeights(
-  {
-    scanFromSrc,
-    scanFromDest,
-    home,
-  }: {
-    scanFromSrc: number | null;
-    scanFromDest: number | null;
-    home: string;
-  },
+  { scanFromSrc, scanFromDest, home }: ResolveHeightsParams,
   logger: Logger
 ): RelayedHeights | null {
   if (!scanFromSrc && scanFromDest) {
--- a/src/binary/ibc-relayer/commands/start.ts
+++ b/src/binary/ibc-relayer/commands/start.ts
@@ -7,6 +7,7 @@ import { Logger } from 'winston';
 import { Link } from '../../../lib/link';
 import { RelayedHeights } from '../../../lib/link';
 import { lastQueriedHeightsFile, registryFile } from '../../constants';
+import { InvalidOptionError } from '../../exceptions/InvalidOptionError';
 import { LoggerFlags } from '../../types';
 import { loadAndValidateApp } from '../../utils/load-and-validate-app';
 import { loadAndValidateRegistry } from '../../utils/load-and-validate-registry';
@@ -15,7 +16,6 @@ import { resolveHomeOption } from '../../utils/options/shared/resolve-home-optio
 import { resolveKeyFileOption } from '../../utils/options/shared/resolve-key-file-option';
 import { resolveMnemonicOption } from '../../utils/options/shared/resolve-mnemonic-option';
 import { signingClient } from '../../utils/signing-client';
-import { InvalidOptionError } from '../../exceptions/InvalidOptionError';
 
 function resolveHeights(
   {
--- a/src/binary/ibc-relayer/commands/start.ts
+++ b/src/binary/ibc-relayer/commands/start.ts
@@ -1,10 +1,12 @@
+import fs from 'fs';
 import path from 'path';
 
 import { sleep } from '@cosmjs/utils';
 import { Logger } from 'winston';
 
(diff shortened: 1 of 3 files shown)


src/binary/ibc-relayer/commands/start.ts
```diff
@@ -14,6 +17,55 @@ import { resolveKeyFileOption } from '../../utils/options/shared/resolve-key-fil
...
+  const lastQueriedHeightsFilePath = path.join(home, lastQueriedHeightsFile);
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const heights = require(lastQueriedHeightsFilePath);
+    logger.info(
+      `Use last queried heights from ${lastQueriedHeightsFilePath} file.`
+    );
+    return heights;
+  } catch {
+    // log?
```
bartmacbartek, line 63: Should we log at this point? What message would be appropriate? (no heights are specified)

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -14,6 +17,55 @@ import { resolveKeyFileOption } from '../../utils/options/shared/resolve-key-fil
 import { resolveMnemonicOption } from '../../utils/options/shared/resolve-mnemonic-option';
 import { signingClient } from '../../utils/signing-client';
 
+function resolveHeights(
+  {
+    scanFromSrc,
+    scanFromDest,
+    home,
+  }: {
+    scanFromSrc: number | null;
```
ethanfrey, line 26: Can we make this struct some other type, like `LoggerFlags` or `LoopFlags` are?

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -14,6 +17,55 @@ import { resolveKeyFileOption } from '../../utils/options/shared/resolve-key-fil
...
+    scanFromDest: number | null;
+    home: string;
+  },
+  logger: Logger
+): RelayedHeights | null {
+  if (!scanFromSrc && scanFromDest) {
+    throw new InvalidOptionError(
+      `You have defined "scanFromDest" but no "scanFromSrc". Both or none "scanFromSrc" and "scanFromDest" must be present.`
```
ethanfrey, line 34: Actually, I would allow only one.

If both are set, return early as you did.
Otherwise, try to load defaults from the file or just `null`.
If one is set, set just those 2 variables.

But... this can wait to another PR. For now, this seems clear enough

Read the "defaults" from file, or just set to 0.
Then

src/binary/ibc-relayer/commands/start.ts
ethanfrey: Log something like `scanning entire history for packets... this may take some time`

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -154,17 +221,26 @@ async function run(options: Options, logger: Logger) {
   await relayerLoop(link, options, logger);
 }
 
-async function relayerLoop(link: Link, options: LoopOptions, logger: Logger) {
+async function relayerLoop(link: Link, options: Options, logger: Logger) {
   // TODO: fill this in with real data on init
   // (how far back do we start querying... where do we store state?)
-  let nextRelay = {};
+  let nextRelay = options.heights ?? {};
```
ethanfrey, line 227: You can remove the TODO line above

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -154,17 +221,26 @@ async function run(options: Options, logger: Logger) {
...
+    lastQueriedHeightsFile
+  );
 
   const done = false;
   while (!done) {
     try {
       // TODO: make timeout windows more configurable
       nextRelay = await link.checkAndRelayPacketsAndAcks(nextRelay, 2, 6);
 
+      fs.writeFileSync(
```
ethanfrey, line 239: :+1:

ethanfrey (approved):
Looks good. Maybe a little recommended polish before merging.

The larger parse changes is really an edge case and should be a separate issue

src/binary/ibc-relayer/commands/start.ts
bartmacbartek: f4f195a

src/binary/ibc-relayer/commands/start.ts
bartmacbartek: 2fa51eb

I extracted it to params type so it's easier to read, however, this type is not reused anywhere.

PR #123, opened 2021-03-16
Prepare for publishing

### LINKED ISSUE #69: Define binary distribution channel
How do we share with people who will run the relayer?

We could push to npm, we could just share a link to the repo and say "build from source", we could provide a docker image with compiled code. Or maybe there are other ideas.

This issue is to define and document such a process. We can implement it in another issue

bartmacbartek:
Closes #69

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "!**/*.spec.*",
     "!**/*.json",
     "LICENSE",
+    "CHANGELOG.md",
     "README.md"
   ],
   "ava": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,11 +1,8 @@
 {
-  "name": "ts-relayer",
-  "version": "0.0.3",
+  "name": "@confio/relayer",
+  "version": "0.1.0",
   "description": "IBC Relayer in TypeScript",
-  "main": "build/main/index.js",
-  "typings": "build/main/index.d.ts",
-  "module": "build/module/index.js",
```
bartmacbartek, line 7: We don't have an entry point to the relayer (only binaries) so there is no reason to keep those.

willclarktech (approved):

Issue #48, opened 2021-02-17
Add support for Packet Delay (on connection)


ethanfrey:
This one is tricky. We need to allow the option when creating a connection, and read the option when using an existing connection.

If delay=0, then the general behavior we have everywhere works.

If delay > 0, then we must do the following to relay packets.

* SendPacket available at height H on chain A
* Submit A header of H+1 on chain B (to prove H)
* Store proof of packet pointing to this header (H+1)
* *Wait delay seconds after update client*
* Submit the SendPacket proof pointing to header (H+1), not a newer header

This means there is a delay between updating a client and being able to use that client header to prove anything. As a potential protection against byzantine networks.

PR #126, opened 2021-03-17
54/ ibc-setup connections

bartmacbartek:
addresses #54

### DIFF
--- a/src/binary/ibc-setup/commands/channels.ts
+++ b/src/binary/ibc-setup/commands/channels.ts
@@ -2,7 +2,7 @@ import path from 'path';
 
 import { Logger } from 'winston';
 
-import { State } from '../../../codec/ibc/core/channel/v1/channel';
+import { State as ChannelState } from '../../../codec/ibc/core/channel/v1/channel';
 import { registryFile } from '../../constants';
 import { borderLessTable } from '../../utils/border-less-table';
 import { generateMnemonic } from '../../utils/generate-mnemonic';
@@ -60,32 +60,24 @@ export async function channels(flags: Flags, logger: Logger) {
   await run(options, logger);
 }
 
-function assureMnemonic(mnemonic: string | null) {
-  if (!mnemonic) {
-    return generateMnemonic();
-  }
-
-  return mnemonic;
-}
-
-function stateAsText(state: State) {
+function channelStateAsText(state: ChannelState) {
   switch (state) {
-    case State.STATE_CLOSED:
+    case ChannelState.STATE_CLOSED:
       return 'Closed';
 
-    case State.STATE_INIT:
+    case ChannelState.STATE_INIT:
       return 'Init';
 
-    case State.STATE_OPEN:
+    case ChannelState.STATE_OPEN:
       return 'Open';
 
-    case State.STATE_TRYOPEN:
+    case ChannelState.STATE_TRYOPEN:
       return 'Tryopen';
 
-    case State.STATE_UNINITIALIZED_UNSPECIFIED:
+    case ChannelState.STATE_UNINITIALIZED_UNSPECIFIED:
       return 'UninitializedUnspecified';
 
-    case State.UNRECOGNIZED:
+    case ChannelState.UNRECOGNIZED:
     default:
       return 'Unrecognized';
   }
@@ -100,7 +92,7 @@ export async function run(options: Options, logger: Logger) {
     throw new Error(`Chain ${options.chain} not found in ${registryFilePath}.`);
   }
 
-  const mnemonic = assureMnemonic(options.mnemonic);
+  const mnemonic = options.mnemonic ?? generateMnemonic();
 
   const client = await signingClient(chain, mnemonic, logger);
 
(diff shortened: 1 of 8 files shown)


src/binary/ibc-setup/commands/connections.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  await run(options, (logger as unknown) as Logger);
+
+  const match = new RegExp(`${link.endA.connectionID}\\s+Open`);
+  t.assert(consoleLog.getCall(-1).calledWithMatch(match));
+});
+
+// test.serial('logs a message when no connections are found', async (t) => {
+//   //
+// });
```
bartmacbartek, line 68: I would like to write a test for the `no connections found` case, but all tests operate on the same chains, and the connections are already created.

I thought of solutions:
- somehow force this kind of tests to run in the very beginning when the chains are fresh
- create a new chain for this kind of tests

I wouldn't say I like any of those. Maybe I'm missing something, and maybe testing this is not worth at all. Thoughts? @ethanfrey @willclarktech

src/binary/ibc-setup/commands/channels.ts
```diff
@@ -105,7 +92,7 @@ export async function run(options: Options, logger: Logger) {
     throw new Error(`Chain ${options.chain} not found in ${registryFilePath}.`);
   }
 
-  const mnemonic = assureMnemonic(options.mnemonic);
+  const mnemonic = options.mnemonic ?? generateMnemonic();
```
bartmacbartek, line 95: I realized this function does exactly the same as a nullish operator.

src/binary/ibc-setup/commands/connections.spec.ts
willclarktech: A fresh chain with no connections sounds sensible to me.

src/binary/ibc-setup/commands/connections.ts
```diff
@@ -0,0 +1,111 @@
...
+    connectionStateAsText(connection.state),
+  ]);
+
+  if (!connections.length) {
+    logger.info(`No connections found for chain "${options.chain}".`);
+
+    return;
+  }
+
+  const output = borderLessTable([['CONNECTION_ID', 'STATE'], ...connections]);
```
bartmacbartek, line 108: Do we want to expose any other information? Currently, it looks like this:
image

Available data:
image

src/binary/ibc-setup/commands/connections.spec.ts
willclarktech: Compare with our test setup in CosmJS, where we have a number of preset accounts including one that's unused: https://github.com/cosmos/cosmjs/blob/main/scripts/wasmd/README.md#preset-accounts

bartmacbartek: How the preset accounts are related to this? I think listing the connections is account agnostic so we would have to setup a new chain.


Another solution I came up with:
- stub ibc client (don't test against running blockchain).

willclarktech: > How the preset accounts are related to this?

I just present them as an analogy: in that case we need accounts as test fixtures and we include accounts that aren't used to help test certain cases. In this project we need chains as test fixtures and including chains with no connections seems equally helpful.

ethanfrey: Okay, there is a 3-chain test setup ticket, but assuming they all have connections. Please add another issue for "Add simapp chain to CI with no IBC connections" and then mention adding more test cases here when it was added. I would not block merging on this, just note a follow-up.

You could also manually test (once) by starting new CI chains and just running this one test file, with `test.only()` on this test case.

src/binary/ibc-setup/commands/connections.ts
ethanfrey: Those are useful infos.

It would be nice to show the clientId... or actually info on the client (dig down to the chainId given that clientID). But I would add that one later as another issue (TODO: write issue).

delayPeriod should be 0 for all the ones we create (hardcoded somewhere). It is very important to show that if non-zero, but we don't support it yet https://github.com/confio/ts-relayer/issues/48 A follow up issue to allow us to configure the delay on creation (but only use 0 in all CLI) and then showing that value if present would be useful.

ethanfrey: For now, just show Delay and ClientID in new columns and we can improve that later. (Delay should always be 0 here, add a test when we can set it).

src/binary/ibc-setup/commands/connections.spec.ts
bartmacbartek: I created [a new issue](https://github.com/confio/ts-relayer/issues/130) and mentioned it in the comment: e2304ce

src/binary/ibc-setup/commands/connections.ts
bartmacbartek: 682cc17

PR #131, opened 2021-03-25
fix: Support chains with more than 100 active validators

clockworkgr:
getValidatorSet() uses the paged validators() query which has a limit of 100 entries.

Switching to the validatorsAll() query will ensure hashes match on networks with more than 100 active validators such as Cosmos Hub

### DIFF
--- a/src/lib/ibcclient.ts
+++ b/src/lib/ibcclient.ts
@@ -402,7 +402,7 @@ export class IbcClient {
     this.logger.verbose(`Get validator set for height ${height}`);
     // we need to query the header to find out who the proposer was, and pull them out
     const { proposerAddress } = await this.header(height);
-    const validators = await this.tm.validators({ height });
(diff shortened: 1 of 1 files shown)


src/lib/ibcclient.ts
```diff
@@ -402,7 +402,7 @@ export class IbcClient {
     this.logger.verbose(`Get validator set for height ${height}`);
     // we need to query the header to find out who the proposer was, and pull them out
     const { proposerAddress } = await this.header(height);
-    const validators = await this.tm.validators({ height });
+    const validators = await this.tm.validatorsAll( height );
```
webmaster128, line 405:
```suggestion
    const validators = await this.tm.validatorsAll(height);
```

ethanfrey (approved):
Great. We went and added validatorsAll to cosmJS to fix this known issue, just forgot to wire it up. Thanks for finding and fixing this.

Issue #105, opened 2021-03-10
Better Logging Messages


ethanfrey:
If you find some bad message, list it here with how to improve... let's try to get to them all some day.

- [x] ics20.ts `Created connections ${conn} (${client}) <=> ${conn} (${client})` should show the chain-id of both sides
- [x] ibcclient.ts - all the debug messages with binary data inside are unreadable... let's base64 encode it and/or truncate any long arrays, so we can see the essential fields (we can manually strip off fields in ibcclient msg before sending to logger)
- [x] When relaying packets or acks, `verbose` level should show the channel and data being sent (maybe sequence?) - if the data is json/utf8 display as text, otherwise as 0x1234 style hex. This will help to observe ics20 packet success/error codes
- [x] All log messages from ibcclient.ts should be prefixed by the chainId (test on verbose)
- [x] Also please use `console.log` (or other output) and *not* the logger for the intended output of the `ibc-setup` commands: https://github.com/confio/ts-relayer/pull/106#pullrequestreview-610538747

alpe:
> verbose: Starting relayer with options {"src":"local_wasm","dest":"local_simapp","home":"/Users/alex/.ibc-setup","mnemonic":"parrot health month then december purpose turkey trim noodle seat curious beauty","srcConnection":"connection-0","destConnection":"connection-0","poll":60,"maxAgeSrc":86400,"maxAgeDest":86400}

Do not log `mnemonic` . Maybe hash it when needed

PR #134, opened 2021-03-30
105/better logging messages

bartmacbartek:
Addresses #105 

- [x] ics20.ts `Created connections ${conn} (${client}) <=> ${conn} (${client})` should show the chain-id of both sides
- [ ] ibcclient.ts - all the debug messages with binary data inside are unreadable... let's base64 encode it and/or truncate any long arrays, so we can see the essential fields (we can manually strip off fields in ibcclient msg before sending to logger)
- [x] All log messages from ibcclient.ts should be prefixed by the chainId (test on verbose)
- [x] Also please use `console.log` (or other output) and *not* the logger for the intended output of the `ibc-setup` commands: https://github.com/confio/ts-relayer/pull/106#pullrequestreview-610538747


src/binary/create-logger.ts
```diff
@@ -92,3 +96,22 @@ export function createLogger(flags: LoggerFlags) {
...
+    });
+    stringifiedRest = stringifiedRest !== '{}' ? ` ${stringifiedRest}` : '';
+
+    const label = info.label ? ` [${info.label}]` : '';
+
+    info[MESSAGE] = `${info.level}:${label} ${info.message}${stringifiedRest}`;
+
+    return info;
+  })();
+}
```
bartmacbartek, line 117: This is a custom format based on [winston.format.simple](https://github.com/winstonjs/logform/blob/master/simple.js) with an extra, optional `label`.

src/binary/create-logger.ts
```diff
@@ -1,3 +1,7 @@
+import jsonStringify from 'fast-safe-stringify';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { MESSAGE } from 'triple-beam';
```
bartmacbartek, line 4: I pull `fast-safe-stringify` and `triple-beam` to create the custom format (down below), however, we don't have those dependencies declared in our project. They're dependencies of `winston`. This is why I have to make `@ts-ignore` comment because ts compiler cannot find type declarations for `triple-bream`.


I think a good practice would be to not import from not-direct dependencies, however, maybe in this case we can make an exception. I'm not sure. Should we install them? WYT @willclarktech

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -208,8 +208,16 @@ async function run(options: Options, logger: Logger) {
     throw new Error('dest chain not found in registry');
   }
 
-  const nodeA = await signingClient(srcChain, options.mnemonic, logger);
-  const nodeB = await signingClient(destChain, options.mnemonic, logger);
+  const nodeA = await signingClient(
+    srcChain,
+    options.mnemonic,
+    logger.child({ label: srcChain })
```
bartmacbartek, line 214: Create a child logger with metadata overridden, `label` in this case. It results in:
```
verbose: [local_wasm] Get header for height ...
```

We can adjust how the `label` is displayed in [simple format function](https://github.com/confio/ts-relayer/pull/134/files#diff-1ec075639d7551edbb2e6eacfa20415bc8bf6896649951d95da8cd3dc2cf06e3R101). Should we move it before the colon?
```
verbose [local_wasm]: Get header for height ...
```

Or maybe can you think of something better-looking?

src/binary/create-logger.ts
ethanfrey: If we import them directly, let's add them to dependencies in `package.json`

src/binary/ibc-relayer/commands/start.ts
ethanfrey: I think the first version is fine.
Nice to see how this is added in.

ethanfrey: Hmmm... I assume you haven't tested this manually:

`const srcChain = chains[options.src];` is a `Chain` object. 

Either use `{label: options.src}` or `{label: srcChain.chainId}`

ethanfrey (commented):
Good start, but please verify what you pass in as a label

src/binary/create-logger.ts
bartmacbartek: 74ec8a4

src/binary/ibc-relayer/commands/start.ts
bartmacbartek: You're right I didn't test it manually but I was going to before pulling out of the draft.

Anyway, it's fixed now, and additionally, I made the `logger.child` function type-safe: 2427a13

From now, `logger.child` options are restricted to `{ label: string }` while before we could pass anything and make mistakes.
```diff
-- child(options: Object): Logger
++ child(options: { label: string }): Logger
```

bartmacbartek:
I did improve the `ics20` output a bit (95f1b84) but I still don't like it. I need to think of something better. If you have any ideas, let me know, thanks!

.eslintrc.json
```diff
@@ -28,6 +28,7 @@
     "sort-imports": [
       "error",
       { "ignoreDeclarationSort": true, "ignoreCase": true }
-    ]
+    ],
+    "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_" }]
```
bartmacbartek, line 32: Don't warn on unsued variables that start with `_`, e.g.: https://github.com/confio/ts-relayer/pull/134/files#diff-e8c0ad1a88c4869847d33b7631f1c6a11037972cd3ac65ecab3834ef0f0bf617R25

src/binary/ibc-relayer/commands/start.ts
ethanfrey: You could extend it to `Record<string, string>` in the future as well, but I doubt we want complex objects in there.

package.json
```diff
@@ -53,9 +53,11 @@
     "ajv": "^7.1.1",
     "axios": "^0.21.1",
     "commander": "^7.1.0",
+    "fast-safe-stringify": "^2.0.4",
```
ethanfrey, line 56: :+1:

src/binary/create-logger.ts
```diff
@@ -1,8 +1,15 @@
+import jsonStringify from 'fast-safe-stringify';
+import { MESSAGE } from 'triple-beam';
 import winston from 'winston';
 
 import { LoggerFlags } from './types';
 import { resolveOption } from './utils/options/resolve-option';
 
+// Re-export Logger interface with type-safe child method.
+export interface Logger extends Omit<winston.Logger, 'child'> {
```
ethanfrey, line 9: Nice use of types

src/binary/ibc-setup/commands/balances.ts
```diff
@@ -77,9 +78,9 @@ export async function run(options: Options, logger: Logger) {
     .join(os.EOL);
 
   if (!balances) {
-    logger.info('No funds found for default denomination on any chain.');
+    console.log('No funds found for default denomination on any chain.');
     return;
   }
 
-  logger.info(balances);
+  console.log(balances);
```
ethanfrey, line 85: :+1:

src/binary/ibc-setup/commands/ics20.ts
```diff
@@ -136,15 +136,27 @@ async function resolveLink(
       connections.dest,
       logger
     );
-    logger.info(
-      `Used existing connections ${link.endA.connectionID} (${link.endA.clientID}) <=> ${link.endB.connectionID} (${link.endB.clientID})`
+    console.log(
+      `Used existing connections [chain_id: ${link.endA.chainId()}, connection_id: ${
```
ethanfrey, line 140: Good to be explicit, but since connectionID is always `connection-XX` and client_id always `07-tendermint-XX`, not sure if we need all these labels. You can remove them, or merge as is and decide to trim down if it is overwhelming.

src/binary/ibc-setup/commands/ics20.ts
```diff
@@ -136,15 +136,27 @@ async function resolveLink(
...
+      }, client_id: ${link.endB.clientID}]`
     );
     return link;
   }
 
   const link = await Link.createWithNewConnections(nodeA, nodeB);
-  logger.info(
-    `Created connections ${link.endA.connectionID} (${link.endA.clientID}) <=> ${link.endB.connectionID} (${link.endB.clientID})`
+  console.log(
+    `Created connections [chain_id: ${link.endA.chainId()}, connection_id: ${
```
ethanfrey, line 153: Same as above

ethanfrey (approved):
Looks good. One proposed (optional) cleanup on the text output.

src/binary/ibc-relayer/commands/start.ts
bartmacbartek: I see but for now, we have no other usage of child logger than with `label` so I'm going to keep it strict.

bartmacbartek:
I wanted to test before merging, it's all good:
image

PR #135, opened 2021-03-30
Add cosmoshub-4 to registry.yaml

orkunkl:


demo/registry.yaml
```diff
@@ -50,3 +43,11 @@ chains:
     faucet: http://49.12.73.189:8002
     rpc:
       - http://188.34.162.78:26657
+  cosmoshub-4:
+    chain_id: cosmoshub-4
+    prefix: cosmos
+    gas_price: 0.01uatom
```
ethanfrey, line 49: Are you sure this gas price works?
I have used `0.025uatom` as a default for some time. Happy if it is cheaper now.

ethanfrey (approved):
Looks good.
Please verify the gas price works.

demo/registry.yaml
orkunkl: Yeah I made it `0.026uatom` after some tx timeouts. I was not sure what was the default uatom gas price among cosmos hub validators.

demo/registry.yaml
```diff
@@ -46,7 +46,7 @@ chains:
   cosmoshub-4:
     chain_id: cosmoshub-4
     prefix: cosmos
-    gas_price: 0.01uatom
+    gas_price: 0.026uatom
```
ethanfrey, line 49: :+1:

ethanfrey (approved):
Better. Maybe add a second rpc endpoint for docs (even if we only use one)

orkunkl:
I reverted musselnet registry comments

PR #144, opened 2021-03-30
Query unbonding period, rather than 3 week default

### LINKED ISSUE #137: Add support for non-default unbonding periods.
When creating a client, we need to set the unbonding period of the remote chain.

This is [currently hardcoded in link.ts](https://github.com/confio/ts-relayer/blob/93078e4c6d7202cd11c013a4f474fc96e61f7cd8/src/lib/link.ts#L51-L54). We should query the chains for that value if possible, or add it as a config option to the registry.yaml.

I am unsure if this is exposed via the staking module. If there is an answer, let's do this auto-config via query. Otherwise, let's add to registry.yaml and default to the current value if not set.

ethanfrey:
Closes #137 

- [x] Make simapp unbonding period 1 week
- [x] Add ibcclient support for unbonding period query
- [x] Use queried unbonding period when building the client


src/lib/ibcclient.ts
```diff
@@ -1368,13 +1382,18 @@ export interface CreateClientArgs {
...
-  unbondingPeriodSec: number,
-  trustPeriodSec: number
+  trustPeriodSec?: number
 ): Promise<CreateClientArgs> {
   const header = await src.latestHeader();
   const consensusState = buildConsensusState(header);
+  const unbondingPeriodSec = await src.getUnbondingPeriod();
+  if (!trustPeriodSec) {
+    trustPeriodSec = Math.floor((unbondingPeriodSec * 2) / 3);
+  }
```
bartmacbartek, line 1396: I think this is not a real case but if `trustPeriodSec` is explicitly set to `0` then it would result in 2/3 unbonding period. Should we check for undefined instead?

```suggestion
  if (trustPeriodSec === undefined) {
    trustPeriodSec = Math.floor((unbondingPeriodSec * 2) / 3);
  }
```

bartmacbartek (approved):

src/lib/ibcclient.ts
ethanfrey: I will make that clear. `trustPeriodSec` or 0 would make it unusable, but better to be clear with types.

src/lib/ibcclient.ts
```diff
@@ -361,6 +364,17 @@ export class IbcClient {
     return this.tm.commit(height);
   }
 
+  /// Returns the unbonding period in seconds
```
willclarktech, line 367: Probably worth using a `/** comment */` docstring for IDE feedback etc.

src/lib/ibcclient.ts
```diff
@@ -361,6 +364,17 @@ export class IbcClient {
...
+  public async getUnbondingPeriod(): Promise<number> {
+    const params = await this.query.staking.params();
+    const seconds = params?.unbondingTime?.seconds?.toNumber();
+    if (!seconds) {
```
willclarktech, line 371: This also throws for 0, is that intended?

src/lib/ibcclient.ts
```diff
@@ -1368,13 +1382,18 @@ export interface CreateClientArgs {
...
   src: IbcClient,
-  unbondingPeriodSec: number,
-  trustPeriodSec: number
+  trustPeriodSec?: number
 ): Promise<CreateClientArgs> {
   const header = await src.latestHeader();
   const consensusState = buildConsensusState(header);
+  const unbondingPeriodSec = await src.getUnbondingPeriod();
+  if (trustPeriodSec === undefined) {
+    trustPeriodSec = Math.floor((unbondingPeriodSec * 2) / 3);
```
willclarktech, line 1395: This will be 0 for very small `unbondingPeriodSec` values, is that intended?

src/lib/queries/staking.ts
```diff
@@ -0,0 +1,26 @@
...
+import { QueryClientImpl } from '../../codec/cosmos/staking/v1beta1/query';
+import { Params } from '../../codec/cosmos/staking/v1beta1/staking';
+
+export interface StakingExtension {
+  readonly staking: {
+    readonly params: () => Promise<Params | undefined>;
+  };
+}
+
+export function setupStakingExtension(base: QueryClient): StakingExtension {
```
willclarktech, line 12: We already export a setup function and extension type from `@cosmjs/stargate` if you want to use that. It's not as lightweight, obviously.

src/lib/ibcclient.ts
ethanfrey: Yes, 0 unbonding period is either impossible of incompatible with IBC. (You need several minutes at absolute minimum)

src/lib/ibcclient.ts
ethanfrey: These numbers are typically in weeks and the shortest I have ever seen is hours. I don't think less can be used with IBC.

ethanfrey:
Okay, just waiting for @tomtau to let me know if this fixes all his testnet issues before merging

PR #148, opened 2021-04-01
Prefer Long multiply instead of native one

bartmacbartek:
https://github.com/cosmos/cosmjs/pull/750#discussion_r605600104


webmaster128 (approved):
Nice, thanks

ethanfrey (approved):
Better

PR #149, opened 2021-04-02
54/ibc-setup channel

### LINKED ISSUE #54: Add Channel setup functionality

bartmacbartek:
Closes #54

### DIFF
--- a/src/binary/ibc-setup/commands/channel.spec.ts
+++ b/src/binary/ibc-setup/commands/channel.spec.ts
@@ -1,0 +1,69 @@
+import fs from 'fs';
+
+import test from 'ava';
+import sinon from 'sinon';
+
+import { Link } from '../../../lib/link';
+import { TestLogger } from '../../../lib/testutils';
+import { Logger } from '../../create-logger';
+import { signingClient } from '../../utils/signing-client';
+
+import { simappChain, wasmdChain } from './chains';
+import { Options, run } from './channel';
+
+const fsReadFileSync = sinon.stub(fs, 'readFileSync');
+const consoleLog = sinon.stub(console, 'log');
+
+const mnemonic =
+  'enlist hip relief stomach skate base shallow young switch frequent cry park';
+
+const registryYaml = `
+version: 1
+
+chains:
+  local_wasm:
+    chain_id: testing
+    prefix: wasm
+    gas_price: 0.025ucosm
+    rpc:
+      - http://localhost:26659
+  local_simapp:
+    chain_id: simd-testing
+    prefix: cosmos
+    gas_price: 0.025umuon
+    rpc:
+      - http://localhost:26658`;
+
+test.beforeEach(() => {
+  sinon.reset();
+});
+
+test.serial('creates channel for given connections and ports', async (t) => {
+  const logger = new TestLogger();
+
+  const ibcClientSimapp = await signingClient(simappChain, mnemonic);
+  const ibcClientWasm = await signingClient(wasmdChain, mnemonic);
+  const link = await Link.createWithNewConnections(
+    ibcClientWasm,
+    ibcClientSimapp
+  );
+
+  const options: Options = {
+    home: '/home/user',
+    mnemonic,
+    src: 'local_wasm',
+    dest: 'local_simapp',
+    srcConnection: link.endA.connectionID,
+    destConnection: link.endB.connectionID,
+    srcPort: 'transfer',
+    destPort: 'custom',
+    ordered: false,
+    version: 'ics20-1',
+  };
+
+  fsReadFileSync.returns(registryYaml);
+
+  await run(options, (logger as unknown) as Logger);
+
+  t.assert(consoleLog.calledWithMatch(/Created channels/));
+});
--- a/src/binary/ibc-setup/commands/channel.spec.ts
+++ b/src/binary/ibc-setup/commands/channel.spec.ts
@@ -1,11 +1,14 @@
 import fs from 'fs';
+import os from 'os';
 
+import { assert } from '@cosmjs/utils';
 import test from 'ava';
 import sinon from 'sinon';
 
 import { Link } from '../../../lib/link';
 import { TestLogger } from '../../../lib/testutils';
 import { Logger } from '../../create-logger';
+import { indent } from '../../utils/indent';
 import { signingClient } from '../../utils/signing-client';
 
 import { simappChain, wasmdChain } from './chains';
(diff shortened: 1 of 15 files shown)


src/binary/ibc-setup/commands/channel.ts
```diff
@@ -0,0 +1,154 @@
...
+export async function channel(flags: Flags, logger: Logger) {
+  const home = resolveHomeOption({ homeFlag: flags.home });
+  const app = loadAndValidateApp(home);
+
+  const keyFile = resolveKeyFileOption({ keyFileFlag: flags.keyFile, app });
+  const mnemonic = await resolveMnemonicOption({
+    interactiveFlag: flags.interactive,
+    mnemonicFlag: flags.mnemonic,
+    keyFile: keyFile,
+    app,
```
willclarktech, line 52: Let's be consistent with use of shorthand here (ideally via a lint setting).

src/binary/ibc-setup/commands/channel.ts
```diff
@@ -0,0 +1,154 @@
...
+    options.srcConnection,
+    options.destConnection,
+    logger
+  );
+
+  const ordering = options.ordered
+    ? Order.ORDER_ORDERED
+    : Order.ORDER_UNORDERED;
+
+  const channels = await link.createChannel(
```
willclarktech, line 137: How about renaming this method `createChannels`?

willclarktech (approved):
LGTM

src/binary/ibc-setup/commands/channel.ts
ethanfrey: It is only one channel, with 2 endpoints.
Maybe the variable needs to be renamed?

ethanfrey (approved):
Looks good.
No tests? (Or did you test manually?)

src/binary/ibc-setup/commands/channel.ts
bartmacbartek: cd2d00b

I renamed the variable to `channel`, however, in the stdout we still print:
```
Created channels for connections  (...)
```

and that's confusing.

Should we change it?
```
Created channel with endpoints for connections (...)
```

or

```
Created endpoints for connections (...)
```

src/binary/ibc-setup/commands/channel.ts
bartmacbartek: Alright! I added an eslint rule and fixed all the shorthand-possible occurrences.

0545096

bartmacbartek:
> Looks good.
> No tests? (Or did you test manually?)

It was under draft and I was planning to add a test. Here is it: 3ddc5d4

but I also tested manually.

src/lib/testutils.ts
```diff
@@ -33,6 +34,7 @@ export class TestLogger implements Logger {
     this.info = shouldLog ? createSpy(console.info) : createFake();
     this.verbose = shouldLog ? createSpy(console.log) : createFake();
     this.debug = createFake();
+    this.child = () => this;
   }
 }
```
bartmacbartek, line 40: I had to stub the `child` method of logger because we use `TestLogger` in binary tests too. Should we move in under `src` directory? @willclarktech

src/binary/ibc-setup/commands/channel.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    ordered: false,
+    version: 'ics20-1',
+  };
+
+  fsReadFileSync.returns(registryYaml);
+
+  await run(options, (logger as unknown) as Logger);
+
+  t.assert(consoleLog.calledWithMatch(/Created channels/));
+});
```
bartmacbartek, line 69: Not sure what else we could assert here.

Also, is one test case enough?

ethanfrey: One test is fine.

If the return value were easily parseable, I would say to query that port/channel on both sides.
Maybe a good reason to make the output nicely parseable.

src/binary/ibc-setup/commands/channel.ts
ethanfrey: How about this (also making it easy to parse as mentioned below):

```
Created channel:
 <src-chain-id>: <src-port-id>/<src-channel-id> (<src-connection-id>) 
 <dest-chain-id>: <dest-port-id>/<dest-channel-id> (<dest-connection-id>)
```

In other tests you can match a line by chain-id, then easily parse out needed info. Note that channel-id's are theoretically not global identifiers (although in practice I think so), so you should always reference them by port/channel to make them easy to query/find later.

bartmacbartek: 386339c

Great 👍 
image

src/binary/ibc-setup/commands/channel.spec.ts
bartmacbartek: 386339c

src/binary/ibc-setup/commands/channel.spec.ts
```diff
@@ -65,5 +68,33 @@ test.serial('creates channel for given connections and ports', async (t) => {
...
+          '.+: (?<destPort>.+)/(?<destChannel>.+) \\(.+\\)',
+        ]),
+      ].join(os.EOL)
+    )
+  );
+
+  assert(match);
+  assert(match.groups);
+
+  const querySrcChannel = await ibcClientWasm.query.ibc.channel.channel(
```
ethanfrey, line 89: nice queries

src/binary/ibc-setup/commands/channel.ts
```diff
@@ -146,13 +148,13 @@ export async function run(options: Options, logger: Logger) {
...
-  console.log(
-    `Created channels for connections [${link.endA.chainId()}, ${
-      link.endA.connectionID
-    }] <=> [${link.endA.chainId()}, ${link.endA.connectionID}]: ${
-      channel.src.channelId
-    } (${channel.src.portId}) => ${channel.dest.channelId} (${
-      channel.dest.portId
-    })`
-  );
+  const output = [
```
ethanfrey, line 151: Interesting `...indent` shortcut

ethanfrey (approved):
Good stuff. Thanks

PR #150, opened 2021-04-07
Update CosmJS v0.25.0-alpha.2

willclarktech:
Tests are passing, which seems to be good news for https://github.com/cosmos/cosmjs/issues/751
I wasn't sure what to do with the debug log message though.


ethanfrey:
When you request a review, can you take it out of draft first?

If there is text saying "WIP" or "TODO" in the description, I will not merge it. But at least this means "potentially ready to merge minus one TODO in the code"

ethanfrey:
>  I wasn't sure what to do with the debug log message though.

The IbcSendMessage was only used in ts-relayer test code, not in production code. We can definitely live with removing the logger.debug call there

src/lib/ibcclient.ts
```diff
@@ -1339,37 +1336,23 @@ export class IbcClient {
...
-        sourcePort,
-        sourceChannel,
-        sender: senderAddress,
-        token,
-        receiver,
-        timeoutHeight,
-        timeoutTimestamp,
-      }),
-    };
-    this.logger.debug('MsgTransfer', msg);
```
webmaster128, line 1362: Is this log message relevant?

willclarktech: See https://github.com/confio/ts-relayer/pull/150#issuecomment-814881096

PR #153, opened 2021-04-12
140/channels filter by connection

### LINKED ISSUE #140: Provide way to query all channels for a connection
Currently there is not way to determine which channel the relayer is processing. Would be great to see it in app.yaml. Also this could may help when migrating relayer process to another machine.

bartmacbartek:
Closes #140

### DIFF
--- a/src/binary/ibc-setup/commands/channels.spec.ts
+++ b/src/binary/ibc-setup/commands/channels.spec.ts
@@ -1,0 +1,278 @@
+import fs from 'fs';
+import os from 'os';
+
+import test from 'ava';
+import sinon from 'sinon';
+
+import { State as ChannelState } from '../../../codec/ibc/core/channel/v1/channel';
+import { Order } from '../../../codec/ibc/core/channel/v1/channel';
+import { ChannelPair, Link } from '../../../lib/link';
+import { TestLogger } from '../../../lib/testutils';
+import { Logger } from '../../create-logger';
+import { signingClient } from '../../utils/signing-client';
+
+import { simappChain, wasmdChain } from './chains';
+import { channelStateAsText, Options, run } from './channels';
+
+const fsReadFileSync = sinon.stub(fs, 'readFileSync');
+const consoleLog = sinon.stub(console, 'log');
+
+const mnemonic =
+  'enlist hip relief stomach skate base shallow young switch frequent cry park';
+
+const registryYaml = `
+version: 1
+
+chains:
+  local_wasm:
+    chain_id: testing
+    prefix: wasm
+    gas_price: 0.025ucosm
+    rpc:
+      - http://localhost:26659
+  local_simapp:
+    chain_id: simd-testing
+    prefix: cosmos
+    gas_price: 0.025umuon
+    rpc:
+      - http://localhost:26658`;
+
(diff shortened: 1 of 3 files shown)


src/binary/ibc-setup/commands/channels.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+test.before(async () => {
+  const ibcClientSimapp = await signingClient(simappChain, mnemonic);
+  const ibcClientWasm = await signingClient(wasmdChain, mnemonic);
+  link = await Link.createWithNewConnections(ibcClientWasm, ibcClientSimapp);
+  channel = await link.createChannel(
+    'A',
+    'transfer',
+    'custom',
+    Order.ORDER_UNORDERED,
+    'ics20-1'
```
willclarktech, line 56: Are these values not available from test utils?

src/binary/ibc-setup/commands/channels.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+    chain: 'local_wasm',
+    port: null,
+    connection: null,
+    mnemonic: null,
+    home: '/home/user',
+  };
+
+  fsReadFileSync.returns(registryYaml);
+
+  await run(options, (logger as unknown) as Logger);
```
willclarktech, line 73: Hm, maybe we should unify the `Logger` types in `lib` and `binary`.

src/binary/ibc-setup/commands/channels.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  const output = consoleLog.getCall(-1).args[0] as string;
+
+  const channelWithDifferentPort = output
+    .split(os.EOL)
+    .slice(1) // remove table head
+    .find((value) =>
+      value.match(
+        new RegExp(
+          `[^\\s]+\\s+(?!${options.port})[^\\s]+\\s+[^\\s]+\\s+[^\\s]+`
```
willclarktech, line 147: Hm, this is a little hard to understand and seems easy to make a mistake and get false negatives here. How about checking that every channel has the correct port?

willclarktech (approved):
Looking good.

src/binary/ibc-setup/commands/channels.spec.ts
bartmacbartek: Yes, I already spotted this here: https://github.com/confio/ts-relayer/pull/149#discussion_r608089104

Created a task: https://github.com/confio/ts-relayer/issues/155

src/binary/ibc-setup/commands/channels.spec.ts
bartmacbartek: You're right. I inverted it and also spotted an edge case with last empty line.

I think it's easier to follow now, see: 281a400

src/binary/ibc-setup/commands/channels.spec.ts
bartmacbartek: I didn't know but there's something. I had to flip things up but it works: https://github.com/confio/ts-relayer/pull/153/commits/c47a93696d3dcf62f3c4c403c3f77f7b9ca0f34a

PR #157, opened 2021-04-14
124/npmignore

### LINKED ISSUE #124: Add testutils to npmignore
Since we removed the `.spec` from the filename.

bartmacbartek:
Closes #124 

- ignore `testutils.ts` and other `*.spec.*` files that were previously included in the published package (reduce package size from `677.8 kB` to `343.8 kB`)
- clean up the `tsconfig.json` and keep only `commonjs` build for now (only `commonjs` was utilized before)
- remove redundant dependencies (leftovers from the initial commit):
  - `cz-conventional-changelog`
  - `cspell`
  - `gh-pages`
  - `typedoc`
- remove redundant npm scripts


.gitignore
```diff
@@ -1,9 +1,8 @@
 .idea/*
+.vscode
```
willclarktech, line 2: Not sure how I feel about this. Some of these files are clearly intended to be included in VCS, but I haven't found them useful in this repo.

willclarktech (approved):

willclarktech:
> reduce package size from 677.8 kB to 343.8 kB

🧑‍🍳 👌

.gitignore
bartmacbartek: You can have whatever you want on your local machine but when it comes to sharing file-level configuration we have cross-editor solution: https://github.com/confio/ts-relayer/blob/main/.editorconfig

willclarktech: Yeah I'm a big fan of editorconfig, but it doesn't really cover the same things here, eg:
- recommended extensions only makes sense if you share the file with others, eg via committing it to git (but I personally found it annoying that VSCode would recommend them every time)
- in my python/poetry repos I've found it useful to set the python path to the local installation in `settings.json` (but not so relevant for Node/npm repos)

PR #158, opened 2021-04-14
Update README.md

### LINKED ISSUE #141: Make README useful as a quick start guide
Move the current README to DEVELOPING.md or such name.

A new README should be geared to new users that want to try it out. It should link prominently to `DEVELOPING.md` as how to build from source and test it out (contributing?) and it should also link to spec/index.md to explain all the details of the commands.

The main focus should be:

* Install from npm (demo stable and nightly)
* Setup a new connection / channel on 2 existing testnets with faucets already in `registry.yaml`. Run relayer and transfer tokens via cli tools
* Show more advanced startup, for network with no faucet (cosmos hub?) to a custom net - this should let anyone connect their network (not in default registry.yaml) to the cosmos hub. It can assumed they figured out the first one and this should explain an advanced case (See https://github.com/confio/ts-relayer/issues/108 maybe we want different flags than those, but this is the use case)
* FAQ to debug simple questions, link to support channels (cosmwasm discord?)

Ethan: mention requirements of nodes (Cosmos SDK 0.41.1+, etc)

bartmacbartek:
Closes #141 
view it here: https://github.com/confio/ts-relayer/blob/9b276cdd2c3be3eb16b05a13b06f1814d5e18a2d/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ Reads the configuration and starts relaying packets.
      - [What is registry.yaml?]()
    - funds addresses on both sides so relayer can pay the fee while relaying packets
 
+   > **NOTE:** Test blockchains `relayer_test_1` and `relayer_test_2` are  running in the public. You do not need to start any blockchain locally to complete the quick start guide.
+
    > **NOTE:** Run `ibc-setup balances` to see the amount of tokens on each address.
 
 2. Create `ics20` channel
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ enormous amount of disk space. You can also trim it to 3 weeks of data with
 `--pruning=custom --pruning-keep-recent=362880 --pruning-keep-every=0 --pruning-interval=100`, which has been
 tested. It is likely you could reduce `pruning-keep-recent` to as low as, say, 3600, but that would need testing.
 
+## Specification
+[Refer to the specification page.](spec/index.md)
 ## Installation
 
 ### NPM
@@ -78,9 +80,7 @@ Reads the configuration and starts relaying packets.
 
    - creates relayer's home directory at `~/.ibc-setup`
    - creates `app.yaml` inside relayer's home with `src`, `dest` and newly generated `mnemonic`
-     - [What is app.yaml?]()
    - pulls default `registry.yaml` to relayer's home
-     - [What is registry.yaml?]()
    - funds addresses on both sides so relayer can pay the fee while relaying packets
 
    > **NOTE:** Test blockchains `relayer_test_1` and `relayer_test_2` are  running in the public. You do not need to start any blockchain locally to complete the quick start guide.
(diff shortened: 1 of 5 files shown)


ethanfrey (approved):
This looks good as a start.

I just pushed some notes on full node / blockchain requirements so we don't get bug reports for incompatible chains. It was too long to include in a comment. But please adjust those as needed (it is good to give them as a warning for anyone using the relayer on a real net, but maybe not before the quick start guide)

README.md
```diff
@@ -1,209 +1,135 @@
...
-# normal flow
-yarn build
-yarn test
-yarn fix
+1. Make sure `wasmd` binary is installed on your system
 
-# advanced checks
-yarn cov
-yarn doc
+   - visit [https://github.com/CosmWasm/wasmd](https://github.com/CosmWasm/wasmd) and follow quick start instructions
```
seantking, line 107: Maybe we could make this part more explicit. I wasn't too sure what is the `quickest` thing to do here. Should I use the dev server specified in the docs and start a chain on `:26657` & `16657` ->  https://github.com/CosmWasm/wasmd#dev-server ?

ethanfrey: It is good to link this as a reference, but we could just be more explicit here. Maybe add this:

1. Ensure you have Go 1.15 installed on your system and that you are running Linux or OSX on amd64 (not arm64 / Mac M1).
2. Ensure that your PATH is set up to include Go binaries  (normally `~/go/bin`)

```
git clone https://github.com/CosmWasm/wasmd.git
cd wasmd
git checkout v0.15.1
make install
```

ethanfrey: And we can update to v0.16.0 when that is ready

bartmacbartek: @seantking The `wasmd` is used here only to generate new accounts and send transactions. There is no need to run any chains locally to complete the quick start.

I added a note about this and improved the `wasmd` installation part: 799b1b3

bartmacbartek: https://github.com/confio/ts-relayer/blob/799b1b3faae1d4f0bb2162f5ef2164956feae5fb/README.md

seantking: Nice, makes sense. :+1:  I think having a local quickstart somewhere would be beneficial also.

spec/config.md
```diff
@@ -34,6 +34,8 @@ chains:
       - gas_price: 0.1umayo
       # the path we use to derive the private key from the mnemonic
       - hd_path: 44'/108'/0'/1'
+      # if you include an optional faucet, it will load the relayer with tokens in `ibc-setup init`
+      faucet: https://faucet.musselnet.cosmwasm.com
```
ethanfrey, line 38: This should be `- faucet` correct?

ethanfrey (approved):
Looks good.

spec/config.md
bartmacbartek: 639a121

PR #159, opened 2021-04-15
Shorten logger.debug output by encoding uint8arrays to base64

### LINKED ISSUE #82: log.debug is unusable in integration tests
It dumps out way too much info and all the binary data is encoded like `[34, 72, 143, ...]`

Let us:
 (1) print this out in base64 in TestLogger (and see if this effects winstonjs as well)
 (2) allow enabling info, verbose, or debug levels in test logger

Update: I made a workaround that debug is never logger in TestLogger. verbose or more only.

We still need to make debug usable. I think pre-processing the msgs before writing to logs would be better than just dumping the message and hoping base64 makes it readable

bartmacbartek:
Closes #82


src/lib/ibcclient.ts
```diff
@@ -27,6 +27,7 @@ import {
   Tendermint34Client,
 } from '@cosmjs/tendermint-rpc';
 import { arrayContentEquals, assert, sleep } from '@cosmjs/utils';
+import cloneDeep from 'lodash/cloneDeep';
```
bartmacbartek, line 30: Is it a crime to bring `lodash` only for one function?

ethanfrey: Not sure about a crime. If we didn't already have >1 MB dependencies, I would say it is too heavy. But I'll leave the call to will here

src/lib/ibcclient.ts
```diff
@@ -1176,7 +1199,18 @@ export class IbcClient {
       };
       msgs.push(msg);
     }
-    this.logger.debug('MsgRecvPacket(s)', { msgs });
+    this.logger.debug('MsgRecvPacket(s)', {
+      msgs: msgs.map((msg) =>
+        formatMessage(msg, (mutableMsg, encode) => {
```
ethanfrey, line 1204: This seems reasonable - manually choosing the fields to shorten rather than trying to automate it... this lets us customize much more easily.

Also, the use of mutableMsg makes this much easier than an immutible transform (even though Will will grimmace, I think it is the right choice)

ethanfrey (commented):
Nice start. What about all the updateClient messages?

src/lib/ibcclient.ts
bartmacbartek: @willclarktech

src/lib/ibcclient.ts
```diff
@@ -82,6 +83,20 @@ import {
...
+  mutateFn: (
+    deepClonedMessage: T,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    encode: (data: Uint8Array) => any
+  ) => void
+) {
+  const deepClonedMessage = cloneDeep(message);
+  mutateFn(deepClonedMessage, toBase64);
+
+  return deepClonedMessage;
```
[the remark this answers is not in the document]
bartmacbartek: The type is inferred so it's `T` (the same as the input message), however, this is not exactly true because `encode` function changes `UInt8Array`s to `string`s.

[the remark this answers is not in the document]
ethanfrey: Then let's be explicit and return `unknown` or `any` or `Record<string, unknown>`.
Ideally the most restrictive type we can pass to this.logger.debug.

src/lib/ibcclient.ts
```diff
@@ -82,6 +83,20 @@ import {
   toIntHeight,
 } from './utils';
 
+function formatMessage<T>(
+  message: T,
+  mutateFn: (
+    deepClonedMessage: T,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    encode: (data: Uint8Array) => any
```
ethanfrey, line 91: I would actually remove this from here, and leave it inside the messages (maybe we want different ways to encode different binary fields).

src/lib/ibcclient.ts
```diff
@@ -731,7 +746,15 @@ export class IbcClient {
         },
       }),
     };
-    this.logger.debug(`MsgUpdateClient`, updateMsg);
+
+    this.logger.debug(
+      `MsgUpdateClient`,
+      formatMessage(updateMsg, (mutableMsg, encode) => {
+        if (mutableMsg.value.header?.value) {
+          mutableMsg.value.header.value = encode(mutableMsg.value.header.value);
```
ethanfrey, line 754: `s/encode/toBase64/` here would let us remove the `encode` argument above and make this more flexible

ethanfrey (commented):
A little more cleanup requested, but otherwise looks good.

I would just add this approach to the other messages (client and channel handshake, client update) and then ready to merge.

src/lib/ibcclient.ts
bartmacbartek: I used `encode` argument because we could extend it with length check and use hex for short arrays as you suggested.

I agree being explicit is better (more readable), also the function itself is generic now: bf99247

bartmacbartek:
>Nice start. What about all the updateClient messages?
38a5a9b

src/lib/ibcclient.ts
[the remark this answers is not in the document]
bartmacbartek: That's right, better to say "I don't know" than lie about types :P Changed it here: https://github.com/confio/ts-relayer/pull/159/commits/bf99247f7cf609125b03c73919a9760f41cf8cb0

src/lib/ibcclient.ts
bartmacbartek: addressed in https://github.com/confio/ts-relayer/pull/159/commits/bf99247f7cf609125b03c73919a9760f41cf8cb0

ethanfrey (approved):
One comment to use a helper to remove all those `as any` casts and make this easier to read/write.

After that, please merge. Will can always review after the fact as we will have some follow up PRs anyway

src/lib/ibcclient.ts
webmaster128: I'm sure Will has a `reduce` command for that

src/lib/ibcclient.ts
willclarktech: Haha, I did!

Issue #128, opened 2021-03-17
Auto-generate docs for using the CLI tools


ethanfrey:
Bartek can you elaborate more on this?

- [ ] When I run `ibc-relayer start --help` for example, it doesn't show the default values. It would be good to output that somehow. (This requires changing how we integrate with commander, or swapping out commander)

bartmacbartek:
The "architecture" looks like this:
```
INPUT --> [commander] --> [options-resolver] --> [run] --> OUTPUT
```

- commander
  - contains command & its options definition
  - parses cli flags
  - generates command help


- options-resolver
  - resolves options based on importance (flag > app file > env etc.)
  - forwards them to `run`

`commander` step is responsible for generating the help message. It's unaware of `options-resolver` step and thus it's omitting some information such as default option value.

We cannot merge `options-resolver` step into `commander` because it contains inter references and async operations.

Moving the default option definition from `options-resolver` to `commander` resolves the problem with the help message, however, it creates another one. If a user doesn't provide a flag then a default value is used and it's going to be picked up in the `options-resolver` step:
```ts
  const poll = resolveOption('poll', { required: true, integer: true })(
    flags.poll, // <-- resolved as first while it should be the last choice
    process.env.RELAYER_POLL,
  );
```

I have a solution to this but IMO it gives too much room for an error:
```ts
  const poll = resolveOption('poll', { required: true, integer: true })(
    notDefault(flags.poll), // <-- wrapping with notDefault returns a value only if provided by an user
    process.env.RELAYER_POLL,
    flags.poll, // <-- default value
  );

  function notDefault(...) {
    // checks if option is in process.argv (it means it was provided by an user)
  }
```

The `commander.js` or `yargs` don't provide any metadata with parsed option (e.g. if default is used).

---

The other problem is that currently we define `Flags` type by hand. It must be maintained and can lead to errors:
```ts
const startCommand = program
  .addOption(homeOption)
  .addOption(srcOption)
  .addOption(destOption)
  ...

type Flags = {
  home?: string
  src?: string
  dest?: string
}

/*
change in startCommand has to be applied to Flags type too. they're not
in sync and we don't test the whole options-resolvers. this is not reliable
*/
```

Ideally, I would like to get the type from where the options are defined. I found it's possible with `yargs`: https://github.com/yargs/yargs/blob/master/docs/typescript.md (first two code blocks). However, I have issues applying this technique with multiple commands in one binary (our case).

---

**Summary**
[...]

PR #167, opened 2021-04-21
Expose default values in usage message

bartmacbartek:
Addresses #128

### DIFF
--- a/src/binary/commander-options.ts
+++ b/src/binary/commander-options.ts
@@ -49,14 +49,16 @@ export const destConnection = new Option(
   '--dest-connection <connection>',
   'Destination connection id'
 );
-export const srcPort = new Option(
-  '--src-port <port>',
-  'Source port to create channel'
-);
-export const destPort = new Option(
-  '--dest-port <port>',
-  'Destination port to create channel'
-);
+export const srcPort = (descriptionPostfix = '') =>
+  new Option(
+    '--src-port <port>',
+    `Source port to create channel${descriptionPostfix}`
+  );
+export const destPort = (descriptionPostfix = '') =>
+  new Option(
+    '--dest-port <port>',
+    `Destination port to create channel${descriptionPostfix}`
+  );
 
 export const addLoggerOptionsTo = (command: commander.Command) => {
   return command
--- a/src/binary/ibc-relayer/commands/start.ts
+++ b/src/binary/ibc-relayer/commands/start.ts
@@ -108,14 +108,12 @@ type Options = {
 } & LoopOptions;
 
 // some defaults for looping
-const defaultOptions: LoopOptions = {
+export const defaults = {
   // check once per minute
   poll: 60,
   // once per day: 86400s
   maxAgeSrc: 86400,
   maxAgeDest: 86400,
-
-  once: false,
 };
 
 export async function start(flags: Flags, logger: Logger) {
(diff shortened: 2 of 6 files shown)


src/binary/ibc-relayer/commands/start.ts
```diff
@@ -108,14 +108,12 @@ type Options = {
 } & LoopOptions;
 
 // some defaults for looping
-const defaultOptions: LoopOptions = {
+export const defaults = {
```
willclarktech, line 111: Why have we lost the type here?

src/binary/ibc-relayer/index.ts
```diff
@@ -40,15 +40,15 @@ const startCommand = program
   .addOption(destConnection)
   .option(
     '--poll <frequency>',
-    'How many second we sleep between checking for packets'
+    `How many second we sleep between checking for packets (default: ${startDefaults.poll})`
```
willclarktech, line 43: Typo: "seconds"

src/binary/commander-options.ts
```diff
@@ -49,14 +49,16 @@ export const destConnection = new Option(
...
-  'Source port to create channel'
-);
-export const destPort = new Option(
-  '--dest-port <port>',
-  'Destination port to create channel'
-);
+export const srcPort = (descriptionPostfix = '') =>
+  new Option(
+    '--src-port <port>',
+    `Source port to create channel${descriptionPostfix}`
```
willclarktech, line 55: What's the `descriptionPostfix` for?

src/binary/ibc-relayer/index.ts
```diff
@@ -40,15 +40,15 @@ const startCommand = program
...
   )
   .option(
     '--max-age-src <seconds>',
-    'How old can the client on src chain be, before we update it'
+    `How old can the client on src chain be, before we update it (default: ${startDefaults.maxAgeSrc})`
```
ethanfrey, line 47: Nice to add these defaults

src/binary/ibc-setup/index.ts
```diff
@@ -59,8 +59,8 @@ const ics20Command = program
   .addOption(srcTrust)
   .addOption(destTrust)
   .addOption(mnemonicOption)
-  .addOption(srcPort)
-  .addOption(destPort)
+  .addOption(srcPort(` (default: ${ics20Defaults.port})`))
```
ethanfrey, line 62: This is nice

ethanfrey (commented):
Looks good to me once you address Will's comments.
(Ping me and I will approve)

src/binary/ibc-relayer/index.ts
bartmacbartek: d30c140

src/binary/commander-options.ts
bartmacbartek: `srcPort` and `destPort` are used by more than one command, however, only one set their default values. I used `descriptionPostfix` to append that information:

```ts
  .addOption(srcPort(` (default: ${ics20Defaults.port})`))
```

src/binary/ibc-relayer/commands/start.ts
bartmacbartek: It just doesn't make sense to me to have this type there.

`LoopOptions` type contains
```
  once: boolean;
```
it can be only true/false (guaranteed by commander) and we don't need to define its default.

Having the type here doesn't give us any benefit because it's still the developer's responsibility to attach those values in the right place.

Also, I want to be consistent with other defaults definition.

ethanfrey (approved):
lgtm

PR #169, opened 2021-04-22
161/stack trace log

### LINKED ISSUE #161: don't dump stack trace on error
Maybe add an eg --trace flag to dump this if you really want (for debugging). Or maybe just enable this with --log-level=debug. Whatever makes the most sense

bartmacbartek:
Closes #161

### DIFF
--- a/spec/logging.md
+++ b/spec/logging.md
@@ -22,6 +22,7 @@ to the console.
 - `--quiet` or `-q` is an alias to `--log-level=error`
 - `--log-file=<path>` - specified a file to write to, logs will be written in JSON format to the file
   (but a human-readable format to the console)
+- `--stack-trace` - prints stack trace along with error messages
 
 Note: if multiple flags that set log-level are set (eg. `-v -q --log-level=info`), then the most permissive level is taken. (In this case `verbose`).
 
--- a/src/binary/create-logger.ts
+++ b/src/binary/create-logger.ts
@@ -79,7 +79,7 @@ export function createLogger(flags: LoggerFlags): Logger {
         handleExceptions: true,
         format: winston.format.combine(
           winston.format.colorize(),
-          simpleFormat()
+          simpleFormat(flags.stackTrace)
         ),
       }),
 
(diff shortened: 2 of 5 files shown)


spec/logging.md
```diff
@@ -22,6 +22,7 @@ to the console.
 - `--quiet` or `-q` is an alias to `--log-level=error`
 - `--log-file=<path>` - specified a file to write to, logs will be written in JSON format to the file
   (but a human-readable format to the console)
+- `--stack-trace` - prints stack trace along with error messages
```
bartmacbartek, line 25: I did `--stack-trace` instead of `--trace` because I saw it a few times already, however, I'm not sure which one is better. WYT?

ethanfrey: Either one is fine. It is to be used infrequently by devs, it doesn't have to be short. `--stack-trace` is probably better

src/binary/create-logger.ts
```diff
@@ -101,15 +101,18 @@ export function createLogger(flags: LoggerFlags): Logger {
...
-      ...info,
-      level: undefined,
-      message: undefined,
-      label: undefined,
-    });
-    stringifiedRest = stringifiedRest !== '{}' ? ` ${stringifiedRest}` : '';
+    let stringifiedRest = '';
+    if (stackTrace) {
+      stringifiedRest = jsonStringify({
+        ...info,
```
ethanfrey, line 109: This doesn't seem to do what I want (I need to compile and run).

I expect extra info (like `{msgs}`) to be passed along normally. Just not the stack trace from error messages. Can you manually test this? I think there is a different place that needs adjusting (probably in the error capturing function)

ethanfrey (changes requested):
This doesn't look like it does what it is supposed to do. Can run it tomorrow

src/binary/create-logger.ts
bartmacbartek: Ok, I got this now: 493bf08

ethanfrey (approved):
Looks better, I need to test this out.

ethanfrey (approved):
Looks good

```
$ ./build/binary/ibc-setup/index.js init --home ~/foobar
error: "src" option is required.

$ ./build/binary/ibc-setup/index.js init --home ~/foobar --stack-trace
error: "src" option is required. {"stack":"Error: \"src\" option is required.\n    at /home/ethan/js/ts-relayer/build/binary/utils/options/resolve-option.js:15:23\n    at /home/ethan/js/ts-relayer/build/binary/ibc-setup/commands/init.js:33:15\n    at Generator.next (<anonymous>)\n    at /home/ethan/js/ts-relayer/build/binary/ibc-setup/commands/init.js:8:71\n    at new Promise (<anonymous>)\n    at __awaiter (/home/ethan/js/ts-relayer/build/binary/ibc-setup/commands/init.js:4:12)\n    at init (/home/ethan/js/ts-relayer/build/binary/ibc-setup/commands/init.js:29:12)\n    at Object.<anonymous> (/home/ethan/js/ts-relayer/build/binary/utils/logger-with-error-boundary.js:18:19)\n    at Generator.next (<anonymous>)\n    at /home/ethan/js/ts-relayer/build/binary/utils/logger-with-error-boundary.js:8:71"}
```

Maybe we should print this stack trace a bit nicer in the cases we show it (but I guess we can't have multi-line logs

PR #170, opened 2021-04-23
Lock all the dependencies at specific version (remove carets)

bartmacbartek:
There was a breaking change introduced between `@cosmjs/*@0.25.0-alpha.2` and `@cosmjs/*@0.25.0-alpha.3` in arguments forwarding. Because of the caret (`^`) all the new consumers download `0.25.0-alpha.3` which is breaking the relayer.

For now, I'm going to lock everything at very specific version, this is a good practice for libraries.


package.json
```diff
@@ -28,25 +28,25 @@
...
-    "axios": "^0.21.1",
-    "commander": "^7.1.0",
-    "fast-safe-stringify": "^2.0.4",
-    "js-yaml": "^4.0.0",
-    "lodash": "^4.17.21",
-    "protobufjs": "~6.10.2",
-    "table": "^6.0.7",
-    "triple-beam": "^1.3.0",
-    "winston": "^3.3.3"
+    "@cosmjs/crypto": "0.25.0-alpha.2",
```
ethanfrey, line 31: I definitely support locking all -alphaX pre-releases. They always have breaking changes between them.

package.json
```diff
@@ -28,25 +28,25 @@
...
+    "@cosmjs/encoding": "0.25.0-alpha.2",
+    "@cosmjs/faucet-client": "0.25.0-alpha.2",
+    "@cosmjs/math": "0.25.0-alpha.2",
+    "@cosmjs/proto-signing": "0.25.0-alpha.2",
+    "@cosmjs/stargate": "0.25.0-alpha.2",
+    "@cosmjs/stream": "0.25.0-alpha.2",
+    "@cosmjs/tendermint-rpc": "0.25.0-alpha.2",
+    "@cosmjs/utils": "0.25.0-alpha.2",
+    "ajv": "7.1.1",
```
ethanfrey, line 40: However, post-1.0 packages should be safe with `^` (or just `~` if you prefer).

ethanfrey (approved):
I would re-add the `^` to all the post-1.0 deps (and `~` to non-alpha pre-1.0)

Also, I am surprised the `yarn.lock` file was ignored, but that happens with `npm install` or I guess `npx`. Shall we also add a `package-lock.json` or are there major issues combining yarn and npm lock files? (That can be another PR)

bartmacbartek:
> I would re-add the `^` to all the post-1.0 deps (and `~` to non-alpha pre-1.0)
> 
> Also, I am surprised the `yarn.lock` file was ignored, but that happens with `npm install` or I guess `npx`. Shall we also add a `package-lock.json` or are there major issues combining yarn and npm lock files? (That can be another PR)

`yarn.lock` or `package-lock.json` files are never published to npm registry, so when installing from registry it doesn't matter if `yarn` or `npm` is used. The lock files are only relevant for installation from the source.

We could use [shrinkwrap file](https://docs.npmjs.com/cli/v6/configuring-npm/shrinkwrap-json) that can be published but it has [some issues with yarn (read last section)](https://classic.yarnpkg.com/en/docs/migrating-from-npm/) so unless we migrate to `npm` as our primary package manager we shouldn't be using that.

In an ideal world, post-1.0 dependencies should support semver. Even if they do, people tend to make mistakes and mark breaking changes as patches or minors. I believe it's best to lock all the dependencies at specific versions to ensure the same deps on every install.

PR #171, opened 2021-04-26
98/drop src and dest

### LINKED ISSUE #98: Revisit spec of app.yaml and cli interaction
app.yaml is a nice convenience and allows us to store all state.

A number of script will "auto-update" it for you as well. Which can lead to undesired behavior if you are using some values from CLI and some from app.yaml.

This was brought up a few times in comments:

* https://github.com/confio/ts-relayer/pull/95/files#r590706164
* https://github.com/confio/ts-relayer/pull/95/files#r590707249
* https://github.com/confio/ts-relayer/pull/94#discussion_r589218696

We need to rethink the approach and adjust the spec. Then update the implementations to match.

bartmacbartek:
Closes #98


ethanfrey (approved):
Looks good.

PR #173, opened 2021-04-27
133/prometheus

### LINKED ISSUE #133: Add prometheus support (local dev)
See https://github.com/confio/ts-relayer/blob/main/spec/metrics.md

Let's discuss the desired approach and metrics to collect before starting on this issue.

@alpe and @orkunkl I would appreciate it if you could add any references here to how you would connect a prometheus server and/or which metrics in particular you would like to collect.

bartmacbartek:
Closes #133


prometheus.yaml
```diff
@@ -0,0 +1,10 @@
+global:
+  scrape_interval: 15s
+  evaluation_interval: 15s
+
+scrape_configs:
+  - job_name: confio_relayer
+
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['host.docker.internal:26660']
```
bartmacbartek, line 10: This is an example configuration file. Is this the right place to store it?

src/binary/ibc-relayer/setup-prometheus.ts
```diff
@@ -0,0 +1,65 @@
...
+      `"setupPrometheus" func shouldn't be initialized more than once.`
+    );
+  }
+  initialized = true;
+
+  if (!enabled) {
+    return getStubbedMetrics();
+  }
+
+  client.collectDefaultMetrics({ prefix: 'confio_relayer_' });
```
bartmacbartek, line 46: Not sure if we want to collect these? Reference:
- https://github.com/siimon/prom-client#default-metrics
- https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -262,5 +294,7 @@ async function relayerLoop(link: Link, options: Options, logger: Logger) {
     logger.info(`Sleeping ${options.poll} seconds...`);
     await sleep(options.poll * 1000);
     logger.info('... waking up and checking for packets!');
+
+    metrics?.pollCounter?.inc();
```
bartmacbartek, line 298: With nullish coalescing operator, we can collect metrics nicely w/o condition blocks (if metrics are disabled).

src/binary/ibc-relayer/setup-prometheus.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Logger } from '../create-logger';
+
+let initialized = false;
+
+function getMetrics() {
+  return {
+    pollCounter: new client.Counter({
+      name: 'poll_counter',
+      help: 'Poll counter',
+    }),
```
bartmacbartek, line 14: This is only an example of custom metric that should be removed. Let's replace this with some real ones.

src/binary/ibc-relayer/setup-prometheus.ts
```diff
@@ -0,0 +1,58 @@
...
+  };
+}
+
+export type Metrics = ReturnType<typeof setupPrometheus>;
```
willclarktech, line 18: Why not define the type here and use it in the function definition?

src/binary/ibc-relayer/setup-prometheus.ts
```diff
@@ -0,0 +1,58 @@
...
+export function setupPrometheus({
+  enabled,
+  port,
+  logger,
+}: {
+  enabled: boolean;
```
willclarktech, line 24: Can we make these `readonly`?

src/binary/ibc-relayer/setup-prometheus.ts
```diff
@@ -0,0 +1,58 @@
+import http from 'http';
+
+import client from 'prom-client';
+
+import { Logger } from '../create-logger';
+
+let initialized = false;
```
willclarktech, line 7: Hm, do we need some kind of global state?

bartmacbartek: It's just nice to have to make sure we don't initialize the HTTP server more than once: https://github.com/confio/ts-relayer/pull/173/files#diff-95ad111838746b4a0030c8e37ff66c5de9df11b1a2eab5aa3f59895c9f70016cR28-R32

I thought of implementing a singleton here but since it takes arguments like `enabled` I decided to just throw on second call.

src/binary/ibc-relayer/setup-prometheus.ts
bartmacbartek: Of course! 4d93ebd

src/binary/ibc-relayer/setup-prometheus.ts
bartmacbartek: I agree it's safer than relying on the inference.

af3e18c

README.md
```diff
@@ -136,19 +137,45 @@ ibc-relayer start --home /home/user/relayer_custom_home
...
+
+To enable metrics collection, pass the `--enable-metrics` flag when starting the relayer:
+
+```sh
+ibc-relayer start --enable-metrics
+```
+
+> **NOTE:** Metrics can also be enabled via an environment variable `RELAYER_ENABLE_METRICS=true`, or with an `enableMetrics: true` entry in the `app.yaml` file, as explained in the [config specification](./spec/config.md#configuration).
+
+The `GET /metrics` endpoint will be exposed by default on port `26660`, which you can override with `--metrics-port` flag, `RELAYER_METRICS_PORT` env variable, or `metricsPort` entry in `app.yaml`.
```
ethanfrey, line 165: :+1:

README.md
```diff
@@ -136,19 +137,45 @@ ibc-relayer start --home /home/user/relayer_custom_home
...
+
+### Local setup
+
+1. Start the relayer with metrics enabled
+2. Spin up the Prometheus instance:
+   ```sh
+   docker run -it -v $(pwd):/prometheus -p9090:9090 prom/prometheus --config.file=prometheus.yaml
```
ethanfrey, line 172: Very cool, a default prometheus config committed in repo. Better than any docs.

prometheus.yaml
ethanfrey: Good to commit it to this repo, so it is maintained.

You could put it in a `samples` or `ops` dir. We can add example `systemd` and other configs there as needed.
(Maybe we move it together with the `registry.conf` file?)

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -67,53 +68,57 @@ function resolveHeights(
...
-  home?: string;
-  src?: string;
-  dest?: string;
-  keyFile?: string;
-  mnemonic?: string;
-  srcConnection?: string;
-  destConnection?: string;
-  scanFromSrc?: string;
-  scanFromDest?: string;
+  readonly interactive: boolean;
```
ethanfrey, line 88: Will was here :smile:

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -171,6 +176,20 @@ export async function start(flags: Flags, logger: Logger) {
     flags.scanFromDest,
     process.env.RELAYER_SCAN_FROM_DEST
   );
+  const enableMetrics =
+    flags.enableMetrics ||
+    Boolean(process.env.RELAYER_ENABLE_METRICS) ||
+    app?.enableMetrics ||
+    false;
+  const metricsPort = resolveOption('metricsPort', {
```
ethanfrey, line 184: Very nice.

src/binary/ibc-relayer/commands/start.ts
```diff
@@ -171,6 +176,20 @@ export async function start(flags: Flags, logger: Logger) {
     flags.scanFromDest,
     process.env.RELAYER_SCAN_FROM_DEST
   );
+  const enableMetrics =
```
ethanfrey, line 179: Why don't you use `resolveOption` here like everywhere else?

Ah, there is no support for boolean, only string and integer?
If that is the case, leave it for this PR and make an issue to add boolean support, pointing to this line

src/binary/ibc-relayer/setup-prometheus.ts
```diff
@@ -0,0 +1,60 @@
...
+  enabled,
+  port,
+  logger,
+}: {
+  enabled: boolean;
+  port: number;
+  logger: Logger;
+}): Metrics {
+  if (initialized) {
+    throw new Error(
```
ethanfrey, line 31: This makes sense to find bugs in our code fast.

I was going to suggest to only set `initialized` to true if `enabled` was true, but that means double-registry bugs only surface in production. It is good like this

src/binary/ibc-relayer/setup-prometheus.ts
```diff
@@ -0,0 +1,60 @@
...
+  pollCounter: client.Counter<string>;
+} | null;
+export function setupPrometheus({
+  enabled,
+  port,
+  logger,
+}: {
+  enabled: boolean;
+  port: number;
+  logger: Logger;
```
ethanfrey, line 28: Logger is not yet used. Do you plan to use it in this PR or a direct follow up?
If not, can we remove it until there is a clear use? We can always add it back then

ethanfrey (approved):
Looks good.

A few minor points that can be addressed in a follow up PR. (There are more prometheus tasks)

src/binary/ibc-relayer/setup-prometheus.ts
bartmacbartek: It is used here: https://github.com/confio/ts-relayer/pull/173/files#diff-95ad111838746b4a0030c8e37ff66c5de9df11b1a2eab5aa3f59895c9f70016cR57

or do you think it's better to do console log? I would rather keep the logger to make sure the endpoint is mentioned in the logs.

prometheus.yaml
bartmacbartek: I realized that `init` command has `registry.yaml` hardcoded to main branch, so I cannot move registry under `samples`. https://github.com/confio/ts-relayer/blob/d8a594bcb602a32a0433608bff90f5629322bb6d/src/binary/ibc-setup/commands/init.ts#L49

One way to "fix" this would be to use parametrized link with version:
```ts
`https://raw.githubusercontent.com/confio/ts-relayer/tree/v${packageJson.version}/demo/registry.yaml`
```

But it fixes the future versions only. Everything till `0.1.3` included will break.

👉  For now, I moved `prometheus.yaml` under `demo`. 92287b8

src/binary/ibc-relayer/commands/start.ts
bartmacbartek: Exactly!

https://github.com/confio/ts-relayer/issues/174

src/binary/ibc-relayer/setup-prometheus.ts
alpe: Very good start! Counter makes sense here for this metric.  👍 

Some notes for a future version:
I have noticed though that the naming does not fit the others, yet:
```
confio_relayer_nodejs_gc_duration_seconds_count{kind="minor"} 80
poll_counter 13
```
With a common prefix we can find all relayer metrics easily. Otherwise some context is missing as a prometheus instance is likely not exclusive to the relayer.

No idea if the relayer is always bidirectional but as this would involve 2 chains the semantics of poll is not fully clear to me. You can use `labels` to mark which chain you polled.

alpe (commented):
Good work! 💐  Very happy to see all the system metrics already 🚀

prometheus.yaml
ethanfrey: I think the update to fixed version is good.
And let's keep it in demo for now (marked deprecated), and make the main copy in another directory (final location)

PR #176, opened 2021-05-03
133/grafana

bartmacbartek:


README.md
```diff
@@ -166,14 +166,30 @@ The `GET /metrics` endpoint will be exposed by default on port `8080`, which you
...
 
+#### Grafana
+
+1. Spin up the Grafana instance:
+   ```sh
+   docker run -d --name=grafana -p 3000:3000 grafana/grafana
+   ```
+2. Navigate to [http://localhost:3000](http://localhost:3000) and log in (`admin`/`admin`)
+3. [Create a Prometheus data source](https://prometheus.io/docs/visualization/grafana/#creating-a-prometheus-data-source)
+   > **NOTE:** Use `http://localhost:9090/` as the server URL, and `Browser` as Access method for the local setup.
```
ethanfrey, line 187: Are you sure this works?
If I want to connect one docker container from another, I usually need something like `--networking=host` or use a different name than localhost (one pointing to the host, not the docker guest)

ethanfrey (commented):
I will try this (or happy if someone else wants to test manually).
One point looks odd in the instructions.

Also, it would be nice to have a sample grafana dashboard config (can you save/serialize those?) showing the sample metrics we currently collect

README.md
alpe: `http://localhost:9090` with the `Browser` access method works as the client browser is doing the ping pong between Prometheus and Grafana. That is okish for local installations but must not be used in production for security reasons. 
You can also use `http://host.docker.internal:9090` with access type `Server` to let Grafana do the queries

Please add a note that this setup is for local testing only, to be on the safe side

alpe (approved):
👍  Would be great if you could add a note that the Grafana setup is for local tests only and not production

README.md
ethanfrey: Nice to know. I didn't realize there was browser ping-pong.

ethanfrey: I like the server comment, as that makes more sense for my mental model, but the other one is cool too

bartmacbartek: I also find this (Server method) more concise. There is no need to mention the `Browser` access method, so I changed the note to use the `Server` instead: 1158237

Thanks for bringing this!

bartmacbartek:
> Also, it would be nice to have a sample grafana dashboard config (can you save/serialize those?) showing the sample metrics we currently collect

I'll come up with something like this shortly after creating some custom metrics.

PR #178, opened 2021-05-05
Upgrade cosmjs deps to 0.25.0

### LINKED ISSUE #177: Broken dependency in 0.1.3
Just a heads up, https://github.com/confio/ts-relayer/blob/v0.1.3/package.json#L36 @confio/relayer locks this dependency which includes this relative path reference: https://github.com/cosmos/cosmjs/blob/v0.25.0-alpha.2/packages/stargate/src/multisignature.ts#L6 but since the @cosmjs/amino package is not in the dependencies, things break

we've gone around it by including the required deps locked to the same version in our relayer-using project so they're installed/resolved in one place but an updated release using 0.25 would be welcome

bartmacbartek:
Closes #177


src/lib/utils.ts
```diff
@@ -83,7 +83,10 @@ export function mapRpcPubKeyToProto(
       secp256k1: pubkey.data,
     };
   } else {
-    throw new Error(`Unknown validator pubkey type: ${pubkey.algorithm}`);
+    throw new Error(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      `Unknown validator pubkey type: ${(pubkey as any).algorithm}`
```
willclarktech, line 88: Heads up: we tend to use "Pubkey" as a technical word, rather than "PubKey".

willclarktech, line 88: @webmaster128 Maybe these types should extend a base pubkey which has an algorithm string property?

willclarktech (approved):

src/lib/utils.ts
webmaster128: Since `pubkey` is of type `never` in this code branch and it is only a safety exception, I'd not over-engineer it. Let's just use
```js
throw new Error(`Unknown validator pubkey type`);
````

willclarktech: (Doesn't apply to this line but existing lines I can't comment on.)

Issue #180, opened 2021-05-07
Tx was submitted but was not yet found on the chain


mark-microtick:
Using both the latest release and the latest commti on main (650bccd) cannot setup a channel.

$ ibc-setup ics20
Used existing connections [gaiad-microtick-testnet, connection-1, 07-tendermint-1] <=> [microtick-testnet-rc6, connection-0, 07-tendermint-0]
info: Create channel with sender gaiad-microtick-testnet: transfer => transfer
(node:451751) UnhandledPromiseRejectionWarning: Error: Transaction with ID B92AE5B89D5BF39C63D59A5E22D7272201113FCB991AA155477D2B145FEF0628 was submitted but was not yet found on the chain. You might want to check later.

Correction: The transaction eventually does appear, but not before it times out apparently

mark-microtick:
gaiad version v4.2.1
microtick is built with SDK 0.42.4

ethanfrey:
@webmaster128 this is likely due to the new timeout in cosmjs. 

Mark, there is an issue to make this configurable for shorter and longer block times.

What are your typical block times? Are you using low gas, such that it wait several blocks before anyone broadcasts it? If you blocktime is less than 10s, i would increase gas price and see if that works

mark-microtick:
So I tried again just increasing gas and it doesn't seem to have an effect.  Only increasing the timeout seems to work here.

mark-microtick:
I don't know what I did before to get it to work.  After restarting the testnet, I can't get the 
```
$ ibc-setup ics
```
step to work now, even changing the gas and the timeouts.

webmaster128:
This is probably fixed when upgrading CosmJS to 0.25.3.

PR #181, opened 2021-05-09
Add missing await for run

webmaster128:
This should fix the `UnhandledPromiseRejectionWarning` from #180 by ensuring that `run`'s errors are propagated properly.


ethanfrey (approved):
Nice catch

PR #183, opened 2021-05-19
Add docs for web app

### LINKED ISSUE #142: Demo relayer lib in webapp
The `lib` dir should work in both node and browser (react app).

Make a simple app that maybe replicates some subset of `ibc-relayer start) as a webapp. It can take an ugly form for setup (instead of app.yaml) and maybe has a hardcoded registry file inside you must select from. But it should work between some testnets so we can demo a browser app relaying packets.

Someone else can take this and make something useful with it.

ethanfrey:
Closes #142 

Thanks for providing the app @clockworkgr 

Does this look good to you? https://github.com/confio/ts-relayer/blob/add-link-to-web-usage/README.md#Web-App


clockworkgr:
Updated the web app to use 0.1.4 and https://github.com/confio/ts-relayer/pull/184 updates link line numbers

Other than that, LGTM

PR #187, opened 2021-05-19
[skip ci] Better document hd paths in registry.yaml

### LINKED ISSUE #182: Add gaia hd_path as example to registry.yaml
I'm doing some testing of the new ibc-go changes with the ts-relayer and I'm using this [neat tool](https://github.com/informalsystems/ibc-rs/pull/928) to manage my gaia nodes. It relies on the `gaiad testnet` command which will generate keys using the [SDK fundraiser path](https://github.com/cosmos/cosmos-sdk/blob/377360a391e80787d99bd0a9ce962a18bee8dc77/types/address.go#L46). I'm new to BIP 39 vs BIP 44 so it took me some digging through the code to figure out how to set the hd path correctly. 

I think it'd be useful to add the fund raiser hd path as an example to the registry.yaml. Maybe something like this?

```
# the path we use to derive the private key from the mnemonic
hd_path: m/44'/108'/0'/1'
# default test keyring hd_path for the SDK
# hd_path: m/44'/118'/0'/0/0
```

I'm happy to open a pr. I'm also a bit fuzzy on the exact technical wording to be used here

ethanfrey:
Closes #182 

I'd like to get @colin-axner to review before we merge and close the issue

### DIFF
--- a/demo/registry.yaml
+++ b/demo/registry.yaml
@@ -10,8 +10,7 @@ chains:
     # The path we use to derive the private key from the mnemonic
     # Note: The hd paths shown have no meaningful relationship to the existing chains.
     # It is recommended practice to use a different hd_path than those commonly used for user accounts.
-    # If you want this, look at the cosmoshub example below
-    hd_path: m/44'/108'/0'/1'
+    hd_path: m/44'/1234'/0'/1'
     # If you include an optional faucet, it will load the relayer with tokens in `ibc-setup init`
     faucet: https://faucet.musselnet.cosmwasm.com
     # You can optionally define a default ics20_port that will be used instead of transfer if no flags set
@@ -24,14 +23,14 @@ chains:
     chain_id: testing
     prefix: wasm
     gas_price: 0.025ucosm
-    hd_path: m/44'/108'/0'/2'
+    hd_path: m/44'/1234'/0'/2'
     rpc:
       - http://localhost:26659
   local_simapp:
     chain_id: simd-testing
     prefix: cosmos
     gas_price: 0.025umuon
-    hd_path: m/44'/108'/0'/3'
+    hd_path: m/44'/1234'/0'/3'
     ics20_port: 'custom'
     rpc:
       - http://localhost:26658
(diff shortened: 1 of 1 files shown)


demo/registry.yaml
```diff
@@ -54,7 +57,10 @@ chains:
     chain_id: cosmoshub-4
     prefix: cosmos
     gas_price: 0.026uatom
-    hd_path: m/44'/108'/0'/6'
+    # Default keyring hd_path for the SDK
+    # This is used for example by `gaiad keys` and most CosmJS webapps
+    # Set this is you want to share the relayer account with a CLI account (not recommended)
+    hd_path: m/44'/108'/0'/0/0
```
colin-axner, line 63:
```suggestion
    hd_path: m/44'/118'/0'/0/0
```
I believe the coin type is [118](https://github.com/cosmos/cosmos-sdk/blob/377360a391e80787d99bd0a9ce962a18bee8dc77/types/address.go#L46)

demo/registry.yaml
```diff
@@ -3,17 +3,20 @@ version: 1
...
+    # Bech32 prefix for addresses
     prefix: wasm
-    # this determines the gas payments we make (and defines the fee token)
+    # This determines the gas payments we make (and defines the fee token)
     gas_price: 0.025umayo
-    # the path we use to derive the private key from the mnemonic
+    # The path we use to derive the private key from the mnemonic
+    # Note: The hd paths shown have no meaningful relationship to the existing chains.
+    # It is recommended practice to use a different hd_path than those commonly used for user accounts.
+    # If you want this, look at the cosmoshub example below
```
colin-axner, line 13: The "If you want this" confuses me as I'm not sure what "this" is
```suggestion
```
I think it'd be completely fine to remove the line. The comments below do a great job at indicating what hd_path compatibility is needed for gaia. It also gives the user more indication that the default hd_path for certain cli tools might be in a different format than `m/44'/108'/0'/1'`

colin-axner (approved):
LGTM, see suggestions. Thanks for updating this, I think it provides good direction. The recommendations help as well since I hadn't considered separating relayer vs user accounts as good practice

PR #189, opened 2021-05-25
Added the handling of begin and end block events

### LINKED ISSUE #188: `BeginBlock` and `EndBlock` events handling
### Context
Currently the relayer is able to relay only the packets that can be found by searching for transactions events. However there are some chains (ie. [Band Protocol](https://github.com/bandprotocol/chain)) that rely on the `BeginBlock` and `EndBlock` methods to emit events that would not be otherwise emitted. 

This is the case of Band's `x/oracle` module that uses the `EndBlock` method to aggregate all the results coming from different reporters, execute the on-chain oracle script and then store the result. Once the result is stored, if the request was performed though an IBC connection the `EndBlock` sends an IBC packet to the source chain to make sure they get the result properly without having to perform a new query later on. 

This method however is not supported by the TS Relayer since it uses the following method to get the events: 
```typescript
public async querySentPackets({minHeight, maxHeight}: QueryOpts = {}): Promise<PacketWithMetadata[]> {
    let query = `send_packet.packet_connection='${this.connectionID}'`;
    if (minHeight) {
      query = `${query} AND tx.height>=${minHeight}`;
    }
    if (maxHeight) {
      query = `${query} AND tx.height<=${maxHeight}`;
    }

    const search = await this.client.tm.txSearchAll({ query });
    
    ...
}
```

### Proposed solution
A possible solution could be the one of using the `/block_results?height={height}` endpoint, querying all the blocks that were not parsed previously to get the `begin_block_events` and `end_block_events` properly. This however would mean basing the solution on a high number of queries of an unreliable part of Tendermint. 

### CCs
CCing some people that might be interested. Particularly: 
- @ethanfrey and @orkunkl
- @taobun and @Benzbeeb (Band developers)

RiccardoM:
This PR adds the support for `being_block_events` and `end_block_events`. Closes #188. 

The code has been tested locally on a `Desmos <-> Band` IBC connection and it works properly.

### Note
The code currently relies on a fork of the `cosmjs/tendermint-rpc` package that is available [here](https://github.com/desmos-labs/cosmjs/tree/riccardo/add-block-search/packages/tendermint-rpc). This is done because the NPM `cosmjs` package does not support the required `/block_events` endpoint. I've already opened a PR towards their repo to implement this feature (see cosmos/cosmjs#810). Until that one is merged, however, we need to rely on the fork.


package.json
```diff
@@ -35,7 +35,7 @@
     "@cosmjs/proto-signing": "0.25.3",
     "@cosmjs/stargate": "0.25.3",
     "@cosmjs/stream": "0.25.3",
-    "@cosmjs/tendermint-rpc": "0.25.3",
+    "@cosmjs/tendermint-rpc": "https://gitpkg.now.sh/desmos-labs/cosmjs/packages/tendermint-rpc?riccardo/add-block-search",
```
ethanfrey, line 38: Wow, very nice. I never knew how to import non-tagged releases from lerna monorepos

src/lib/endpoint.ts
```diff
@@ -92,6 +122,18 @@ export class Endpoint {
     return ([] as PacketWithMetadata[]).concat(...resultsNested);
   }
 
+  // returns all packets (auto-paginates, so be careful about not setting a minHeight)
+  public async querySentPackets({
+    minHeight,
+    maxHeight,
+  }: QueryOpts = {}): Promise<PacketWithMetadata[]> {
+    const txsPackets = await this.getPacketsFromTxs({ minHeight, maxHeight });
+    const eventsPackets = await this.getPacketsFromBlockEvents({ minHeight, maxHeight });
```
ethanfrey, line 131: This will throw an unspecified (and unhandled) error if used on any chain less than tendermint v0.34.9 (cosmos-sdk v0.42.4). I would tend to making this more robust, unless there really is no need to handle earlier chains.

See comments on https://github.com/cosmos/cosmjs/pull/810#pullrequestreview-667718860

src/lib/utils.ts
```diff
@@ -96,7 +97,7 @@ export function timestampFromDateNanos(
   const nanos = (date.getTime() % 1000) * 1000000 + (date.nanoseconds ?? 0);
   return Timestamp.fromPartial({
     seconds: new Long(date.getTime() / 1000),
-    nanos,
```
ethanfrey, line 99: Lots of small changes from your editor/prettier config.

Can you run `yarn fix` to bring this into a [standard format](https://github.com/confio/ts-relayer/blob/main/package.json#L10)? (Maybe we need to add an explicit prettier config for the editors.

src/lib/utils.ts
```diff
@@ -133,53 +134,53 @@ export function buildClientState(
...
+      length: LengthOp.VAR_PROTO
     },
     innerSpec: {
       childOrder: [0, 1],
       minPrefixLength: 1,
       maxPrefixLength: 1,
       childSize: 32,
-      hash: HashOp.SHA256,
-    },
+      hash: HashOp.SHA256
```
ethanfrey, line 160: Note to self: I think I exposed these in the most recent ics23 package. I should see if I can import them instead of recreating them here.

src/lib/utils.ts
```diff
@@ -1,9 +1,10 @@
 import { fromUtf8, toHex, toUtf8 } from '@cosmjs/encoding';
 import { BroadcastTxFailure, Coin, logs, StdFee } from '@cosmjs/stargate';
+import { parseEvent } from '@cosmjs/stargate/build/logs';
```
ethanfrey, line 3: Importing from the build dir smells funny.
This is [kind of exposed top-level](https://github.com/cosmos/cosmjs/blob/79396bfaa49831127ccbbbfdbb1185df14230c63/packages/stargate/src/index.ts#L55)

I think you need :

```ts
import { logs } from '@cosmjs/stargate';
const { parseEvent } = logs;
```

src/lib/utils.ts
```diff
@@ -193,6 +194,26 @@ interface ParsedEvent {
...
+    .filter(({ type }) => type === 'send_packet')
+    .map(({ type, attributes }) => ({
+      type,
+      attributes: attributes.map(({ key, value }) => ({
+        key: decodeBase64(key),
+        value: decodeBase64(value)
+      }))
+    }));
+
+  const flatEvents = ([] as ParsedEvent[]).concat(allEvents.map(parseEvent));
```
ethanfrey, line 213: In any case, nice you do `parseEvents` to do all the validity checks

ethanfrey (commented):
Nice contribution. Thank you very much.

It is obviously blocked on https://github.com/cosmos/cosmjs/pull/810 being merged and there are a few cleanup items to do, but in general the code looks good. I need to ensure it doesn't break the CI and ideally would like to add a cases for:

* Querying still working with v0.41.x (this is broken in the PR, suggestions to fix somehow)
* Relaying handles events in a "positive test case" (eg. desmos -> band)

I can take care of setting up the CI to test this, but I would need minimally docker images for both chains as well as a list of setup commands you you to establish this channel and trigger packet creation, so we can verify some endblock transactions properly relayed.

I am fine with doing the CI tests in a separate PR (once both chains are tagged), but it is important to have some test there to prevent future regressions.

ethanfrey:
Oh yeah, you should run `yarn changeset` to add a line about what feature this adds (it can be a patch release)

RiccardoM:
@ethanfrey I should have fixed everything and added the change set as well

ethanfrey (approved):
Looks clean. My main concerns (besides CI integration) have been addressed.

Now waiting the CosmJS PR to be in a tagged release.

ethanfrey:
I was looking to prepare this for a merge now, but I think this is still not in a CosmJS release.

[Your commit](https://github.com/cosmos/cosmjs/commit/961b46d36ef81eb5590b20fce5dd0de20a2a8a66) is in `cosmjs:main`, but I don't see it in [the `v0.25.4` commit history](https://github.com/cosmos/cosmjs/commits/v0.25.4).

 I guess we need to wait til v0.26.0 and update this then?

RiccardoM:
> I guess we need to wait til v0.26.0 and update this then?

I guess so, yeah :cry: Any ETA on this @webmaster128?

webmaster128:
If this is blocking, I'd rather backport the feature to CosmJS 0.25.x than rushing out 0.26.

ethanfrey:
If it can be backported easily, I can update ts-relayer to use it.
I don't know how urgent this is, just one of my pending tasks I wanted to clean up.

@RiccardoM are you using this code now? Or was it just a demo to test compatibility?

RiccardoM:
> @RiccardoM are you using this code now? Or was it just a demo to test compatibility?

We will use this feature in the next couple of weeks most likely. We are ending our implementation inside Desmos, then we will perform an on-chain upgrade to release it. Targeting the first week of July, so it might be worth backporting this

ethanfrey:
Nice, I would love to see you using a proper tagged release of the code. Much less compatibility and maintenance headaches

webmaster128:
At https://github.com/cosmos/cosmjs/pull/839 you find a backport of the change. The known limitation is that block height 1 cannot be decoded. The fix for that requires breaking changes. Please have a look and let me know if that would unblock you.

package.json
webmaster128: Nice. But where is the TS -> JS compiled?

webmaster128:
CosmJS 0.25.5 should serve your needs now

ethanfrey:
@RiccardoM please run `yarn fix:prettier` to fix up the code, so the CI will run it.

If it passes, I can merge this. I will add release notes for 0.2.0 (changing supported versions) and cut a release with this.

ethanfrey:
Okay, failing test now.
This is cuz we use v0.41.1 simd in the tests.

Do you want to update https://github.com/confio/ts-relayer/blob/main/scripts/simapp/env#L4 to v0.42.4 ?

That *should* pass.

RiccardoM:
@ethanfrey I've updated the envs and not it passes. I've also opened a new PR (#197) that I think should be included in the next release as well since Band requires custom gas limits to be set to receive packets properly

package.json
RiccardoM: @webmaster128 I think you can read more [here](https://gitpkg.vercel.app/guide/#advanced-how-this-function-is-implemented). It should be compiled by them

PR #191, opened 2021-06-09
Add home option to ics20 command

davepuchyr:
Fix the `error: unknown option '--home'` error on `ibc-setup ics20` for the case where `ibc-setup init...` was run with the `--home` option.


ethanfrey (approved):
Nice catch.

Can you add a changelog entry as explained here: https://github.com/confio/ts-relayer/blob/main/DEVELOPMENT.md#maintain-a-clean-changelog

Then it will pass CI and I can merge it.

Issue #190, opened 2021-05-26
relayer_test_1 and relayer_test_2 faucets are down


charleenfei:
registry.yaml indicates these are still up, this leads to the `ibc-setup init --src relayer_test_1 --dest relayer_test_2` command from the README erroring out (ECONNREFUSED)

ethanfrey:
Thank you for pointing this out.
We shut down those testnets thinking they were only used for internal testing. Forgot we started using them in the README (oops).

We are looking to add two new testnets that will be long-running and have a working faucet we can point to. It was a bit much to run 2 nets just for the relayer.  Is there a new gaia testnet coming out soon? Anyone else?

We have a new cosmwasm testnet out we could use as one endpoint.

charleenfei:
sorry for the late response on this. noted!

shahankhatch:
Currently, a Gaia testnet isn't being planned for the short-term.

chengwenxi:
@ethanfrey IRISnet nyancat testnet is long-running and you can use it.

Exposed ports: https://github.com/irisnet/testnets/tree/master/nyancat
Faucet: https://discord.gg/X6dZZxs3#nyncat-faucet

ethanfrey:
@chengwenxi 

I just added `nyancat` to the registry and updated the readme (in #193) and I managed to connect to nyancat and the relayer created a connection with oysternet and an ics20 channel as well. 

When looking into the send commands, I realized it is based on https://github.com/irisnet/irishub/tree/v0.16.0 which comes from Nov 2020. We only support Cosmos SDK 0.41.1+ in the relayer and I only see problems trying to support a testnet using a pre-release version of ibc. At least in the initial README file.

I am happy to keep it in the registry, but hope for a more "standard" one for the readme.

chengwenxi:
> When looking into the send commands, I realized it is based on https://github.com/irisnet/irishub/tree/v0.16.0 which comes from Nov 2020. We only support Cosmos SDK 0.41.1+ in the relayer and I only see problems trying to support a testnet using a pre-release version of ibc. At least in the initial README file.

It's https://github.com/irisnet/irishub/tree/v1.1.1, which is based on Cosmos SDK v0.42.4.

View version:
```bash
curl http://35.234.10.84:1317/node_info | jq .application_version
```

ethanfrey:
Nice. I was surprised everything was working so well. This explains it.

[...]

PR #196, opened 2021-06-22
Add iris installation

chengwenxi:
Resolves: https://github.com/confio/ts-relayer/issues/190#issuecomment-864959173


README.md
```diff
@@ -111,33 +111,44 @@ Reads the configuration and starts relaying packets.
...
-3. Create a valid IrisNet address to send tokens to
+4. Create a valid IRISnet address to send tokens to
 
    ```sh
-   RCPT=$(ibc-setup keys list | grep nyancat | cut -d' '  -f2)
+   iris keys add receiver
    ```
 
-   TODO: setup irisnet binary and make real account there
+   [Get testnet tokens](https://github.com/irisnet/testnets/tree/master/nyancat#faucet) if you want to send tokens to `oysternet`.
```
ethanfrey, line 140: Note, you can only use the faucet once per 24 hours, so we cannot fund the relayer and the receiver in the same day. Maybe a comment here on that, or they just get the error message.

ethanfrey (approved):
Thanks for adding this info. Just missing an empty changeset. I will add it

README.md
chengwenxi: Thanks for your note, we will increase the limit to 2 or more. ❤️

PR #197, opened 2021-06-24
Added the ability to specify the gas limits for different chains

RiccardoM:


ethanfrey:
I will review after lunch.

Can you please add a changeset and rebase on main

spec/config.md
```diff
@@ -32,6 +32,28 @@ chains:
       - prefix: wasm
       # this determines the gas payments we make (and defines the fee token)
       - gas_price: 0.1umayo
+      # this defines the gas limits we are going to use
```
ethanfrey, line 35: This no longer exists in the registry.yaml.
We should update this to point to oysternet

src/binary/utils/signing-client.ts
```diff
@@ -30,6 +30,18 @@ export async function signingClient(
   const options: IbcClientOptions = {
     prefix: chain.prefix,
     gasPrice: GasPrice.fromString(chain.gas_price),
+    gasLimits: {
```
ethanfrey, line 33: Looks good. Can you add a test that this works?

Like setting these very low in a test config, and asserting that running a command leads to an error. This would ensure they are actually used.

ethanfrey (commented):
Nice stuff. Two minor comments.

@RiccardoM do you feel comfortable adding the requested test?

ethanfrey:
I will try to update this PR so I can merge and release.

ethanfrey:
Closing in favor of #199 which builds on this with the missing pieces

PR #208, opened 2021-12-01
Upgrade cosmjs 0.27

webmaster128:


src/lib/ibcclient.ts
```diff
@@ -957,8 +957,8 @@ export class IbcClient {
       [msg],
       calculateFee(this.limits.connectionHandshake, this.gasPrice)
```
ethanfrey, line 958: I guess using auto-gas is another issue?
I would love to see that. But I guess one thing at a time.

ethanfrey (approved):
lgtm. just needs a changelog file

PR #209, opened 2021-12-01
Migrate to cosmjs-types

webmaster128:
After #208.

This tries to replace the locally generated types with `cosmjs-types`. Turns out this does not work as long as the JS type `Date` is used for `Timestamp`, which is a conversion that is losing information.


ethanfrey:
Issue seems to be with protobuf compilation flags. We had to use:

`  --ts_proto_opt="esModuleInterop=true,forceLong=long,useOptionals=true,useDate=false" \`

Note `useDate=false`. I had to PR upstream for this to work properly. We need to maintain nanosecond precision for ibc validation. Using date will truncate to milliseconds.

Maybe you try compiling a branch of cosmjs-types with `useDate=false` and see if we can import, then see if you really want to do that?

webmaster128:
Superseded by #212

PR #211, opened 2021-12-07
Upgrade ts-proto

webmaster128:
This upgrades ts-proto and found a bug fixed in https://github.com/confio/ts-relayer/pull/211/commits/3176cc426a5b2fbd0b3d30f6a538570ccbc8d5df. Here it was assumed that an empty pagination key is represented as undefined. But the correct value for no pagination key is an empty byte array.

### DIFF
--- a/src/lib/queries/ibc.ts
+++ b/src/lib/queries/ibc.ts
@@ -282,7 +282,7 @@ export function setupIbcExtension(base: QueryClient): IbcExtension {
             });
             channels.push(...response.channels);
             key = response.pagination?.nextKey;
-          } while (key);
+          } while (key && key.length);
           return {
             channels,
             height: response.height,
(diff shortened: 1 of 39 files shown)


src/lib/queries/ibc.ts
```diff
@@ -307,7 +307,7 @@ export function setupIbcExtension(base: QueryClient): IbcExtension {
             });
             channels.push(...response.channels);
             key = response.pagination?.nextKey;
-          } while (key);
+          } while (key && key.length);
```
ethanfrey, line 310: These are the only changes in non-autogenerated code, right?

ethanfrey (approved):
thanks

src/lib/queries/ibc.ts
webmaster128: Right

PR #212, opened 2021-12-07
Migrate to cosmjs types (2nd attempt)

webmaster128:
After changing the Timestamp type from Date to an object in cosmjs-teypes 0.4 and after #211, the generated types are the same. We can remove them here to avoid unnecessary clutter and reduce compile time.


ethanfrey (approved):
Nice to see all this fully working upstream now. And no more custom proto types.

That said, we should include a link to an older version as a way of registering your own types. (I think this question comes up from time to time)

PR #237, opened 2022-07-30
Updated Discord faucet link

daniel-farina:
New link: https://discord.gg/NbB8zCPX


ethanfrey (approved):
Thank you for this fix.

I will ignore the changes requirement here, but generally use `yarn changeset` to make such a file, so we can aggregate them for releases

Issue #47, opened 2021-02-17
Support Ordered Channels


ethanfrey:
Depends on #46 (blocked)

okwme:
Would love to know more about when this feature might become available. it's a necessary feature for interchain accounts to be ale to use the ts-relayer (cc @seantking)

ethanfrey:
The priority depends on how urgently the interchain accounts will be on the cosmos hub. Or any major mainnet actually.

Maybe we can chat a bit about all the schedule of the various external moving parts (outside of github). It could be the next non-bugfix feature we work on if this will be deployed soon.

PR #239, opened 2022-07-30
Make poll time configurable

ethanfrey:
We were hardcoding some sleeps that made sense for CI.
These may be creating the problems in running on mainnets.
Making them configurable in the registries and using conservative values.


ethanfrey:
@daniel-farina happy if you can check this out (try the code) - you will need to update your registry.yaml file

sascha1337:
can we get ordered timeouts, to use ICA with ts-relayer ? 
like here?
https://github.com/cosmos/relayer/commit/8d5f79207ae9d1dcc8cb3f8210e0f05cb6c39688#diff-ed9e8e86a269a3118fe1634dc1559607ffa6a11e3b4fd6457122b2888c4f4056

ethanfrey:
This is a long Wishlist https://github.com/confio/ts-relayer/issues/47

Not sure when I do it, but happy for a PR. Let's see how much uptake comes with ts-relayer.

PR #245, opened 2022-09-16
Wait for indexer instead of whole block

webmaster128:
Using `waitOneBlock` is very heavy since it always waits a whole blocktime. If we are a bit out of luck with the node, it sends 3 requests per chain and sleeps for 2 blocktimes. But here we don't need a new block but only get indexed for what we have a DeliverTxResponse already.

### DIFF
--- a/src/lib/link.ts
+++ b/src/lib/link.ts
@@ -538,8 +538,8 @@ export class Link {
 
     // let's wait a bit to ensure our newly committed acks are indexed
     await Promise.all([
-      this.endA.client.waitOneBlock(),
-      this.endB.client.waitOneBlock(),
+      this.endA.client.waitForIndexer(),
+      this.endB.client.waitForIndexer(),
     ]);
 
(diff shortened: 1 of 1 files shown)


ethanfrey (approved):
I think I added this in some debugging.

waitOneBlock actually does some checks to ensure (polling and query)

waitForIndexer is a total guess as there is no feedback when such results will be available.

Fair to change it, I would really like to make `waitForIndexer` actually have reasonable semantics

webmaster128:
> waitForIndexer is a total guess as there is no feedback when such results will be available.
> 
> Fair to change it, I would really like to make `waitForIndexer` actually have reasonable semantics

Yeah :/ It has always been that way in all CosmJS tests. A block is committed, height increased. But it takes 25-100ms until we are able to access the data. I think worst case we can increase the value and will still be much more efficient than sending network requests and waiting full block times.

PR #246, opened 2022-09-16
Make RelayedHeights a Record to avoid unsafe casting

webmaster128:
Just type system tuning


webmaster128:
Looking at https://stackoverflow.com/questions/65799316/why-cant-an-interface-be-assigned-to-recordstring-unknown, the problem is more fundamental. Giving up for now.

PR #249, opened 2022-10-02
Version bumps

faddat:
Yarn upgrade

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -25,243 +25,249 @@
   dependencies:
     "@babel/highlight" "^7.10.4"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.16.7":
-  version "7.16.7"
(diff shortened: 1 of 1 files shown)


ethanfrey (commented):
Thanks for the commit.

For the CI, we ensure there are proper changelog entries when we cut a release, following this pattern: https://github.com/confio/ts-relayer/blob/main/DEVELOPMENT.md#maintain-a-clean-changelog

TL;DR: run yarn changeset and enter some info on this (patch level), so we can include it in the next changelog

ethanfrey:
you may also want to rebase on main, but there are no conflicts

webmaster128:
Closing as outdated. I suggest upgrading dependencies individually to avoid conflicts when integrating the work. But for ts-relayer users this has no impact as the lockfile is not publiched to npm. It is just the versions used in development.

PR #250, opened 2022-10-02
Minor bug fix

Anmol1696:
`ibc-setup init` fails due to trailing commas in yaml file which is unable to read

### DIFF
--- a/.changeset/fix-demo-files.md
+++ b/.changeset/fix-demo-files.md
@@ -1,0 +1,5 @@
+---
+'@confio/relayer': patch
+---
+
+Fix demo yamls by removing trailing commas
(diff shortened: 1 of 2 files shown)


Anmol1696:
the CI seems to be broken here

ethanfrey:
Thanks for the commit.

For the CI, we ensure there are proper changelog entries when we cut a release, following this pattern: https://github.com/confio/ts-relayer/blob/main/DEVELOPMENT.md#maintain-a-clean-changelog

TL;DR: run `yarn changeset` and enter some info on this (patch level), so we can include it in the next changelog

Anmol1696:
Ahh my bad. changeset is interesting, will keep this in mind from next time onwards. Thanks

ethanfrey (approved):
thanks for the contribution

PR #253, opened 2022-10-13
Upgrade CosmJS to 0.29.2

webmaster128:
- Upgrade CosmJS to 0.29.2
- Use new TypeScript-friendly `isDefined` to narrow down list type via `.filter`


ethanfrey (approved):
lgtm

Issue #255, opened 2022-10-19
Sending multiple packets from one tansaction (or msg?) not supported


webmaster128:
I recognized issues when multiple packets are sent from one transaction. Turns out the `log` string we get from Tendermint is buggy (see https://github.com/tendermint/tendermint/issues/9595). So most likely the code here is correct, but I want to bring this up to help other users trying to debug their issues.

Some ideas to improve the situation:
- Ensure my bug report in Tendermint is valid and this is not a CosmWasm or wasmd bug
- Use the events array instead of the raw log to work around it

webmaster128:
The place where this pops up is:

```ts
  private async getPacketsFromTxs({
    minHeight,
    maxHeight,
  }: QueryOpts = {}): Promise<PacketWithMetadata[]> {
    let query = `send_packet.packet_connection='${this.connectionID}'`;
    if (minHeight) {
      query = `${query} AND tx.height>=${minHeight}`;
    }
    if (maxHeight) {
      query = `${query} AND tx.height<=${maxHeight}`;
    }

    const search = await this.client.tm.txSearchAll({ query });
    const resultsNested = search.txs.map(({ hash, height, result }) => {
      const parsedLogs = logs.parseRawLog(result.log);
      // we accept message.sender (cosmos-sdk) and message.signer (x/wasm)
      let sender = '';
      try {
        sender = logs.findAttribute(parsedLogs, 'message', 'sender').value;
      } catch {
        try {
          sender = logs.findAttribute(parsedLogs, 'message', 'signer').value;
        } catch {
          this.client.logger.warn(
            `No message.sender nor message.signer in tx ${toHex(hash)}`
          );
        }
      }
      return parsePacketsFromLogs(parsedLogs).map((packet) => ({
        packet,
        height,
        sender,
      }));
    });
    return ([] as PacketWithMetadata[]).concat(...resultsNested);
  }
```

ethanfrey:
> Use the events array instead of the raw log to work around it

This would make it worse most likely, as it would merge the multiple packet events of the same type into one packet. Using raw_log json should keep them as independent events.

Can you attach the JSON out of a raw_log where two ibc packets were emitted in the same tx? It may be a bug in parsing. Having a fixed json to test against would allow us to debug this much quicker

webmaster128:
> This would make it worse most likely, as it would merge the multiple packet events of the same type into one packet. Using raw_log json should keep them as independent events.

[...]

PR #256, opened 2022-10-19
Create test for parsePacketsFromLogs

webmaster128:
This test came up as part of the debugging of https://github.com/confio/ts-relayer/issues/255. Unfortunately I don't have input data for a case with multiple packets. But at least this provides an example that is easy to extend.

### DIFF
--- a/src/lib/utils.spec.ts
+++ b/src/lib/utils.spec.ts
@@ -1,3 +1,5 @@
+import { fromHex } from '@cosmjs/encoding';
+import { logs } from '@cosmjs/stargate';
 import {
   fromRfc3339WithNanoseconds,
   ReadonlyDateWithNanoseconds,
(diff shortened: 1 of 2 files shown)


src/lib/utils.spec.ts
```diff
@@ -8,12 +10,38 @@ import Long from 'long';
...
   heightGreater,
   parseHeightAttribute,
+  parsePacketsFromLogs,
   parseRevisionNumber,
   secondsFromDateNanos,
   timeGreater,
   timestampFromDateNanos,
 } from './utils';
 
+test('parsePacketsFromLogs works for one packet', (t) => {
```
ethanfrey, line 20: Thanks for the test case

src/lib/utils.spec.ts
```diff
@@ -8,12 +10,38 @@ import Long from 'long';
...
..."packet_connection","value":"connection-31"},{"key":"packet_data","value":"{\\"after\\":\\"1666164035856871113\\",\\"sender\\":\\"juno19pam0vncl2s3etn4e7rqxvpq2gkyu9wg2czfvsph6dgvp00fsrxqzjt5sr\\",\\"job_id\\":\\"dapp-1-1666164017\\"}"},{"key":"packet_data_hex","value":"7b226166746572223a2231363636313634303335383536383731313133222c2273656e646572223a226a756e6f313970616d30766e636c32733365746e34653772717876707132676b797539776732637a66767370683664677670303066737278717a6a74357372222c226a6f625f6964223a22646170702d312d31363636313634303137227d"},{"key":"packet_dst_channel","value":"channel-10"},{"key":"packet_dst_port","value":"wasm.nois1j7m4f68lruceg5xq3gfkfdgdgz02vhvlq2p67vf9v3hwdydaat3sajzcy5"},{"key":"packet_sequence","value":"7489"},{"key":"packet_src_channel","value":"channel-42"},{"key":"packet_src_port","value":"wasm.juno1e7vs76markshus39eyfefh2y3t9guge4t0kvqya3q6vamgsejh4q8lxtq9"},{"key":"packet_timeout_height","value":"0-0"},{"key":"packet_timeout_timestamp","value":"1666167632856871113"}]},{"type":"wasm","attributes":[{"key":"_contract_address","value":"juno1e7vs76markshus39eyfefh2y3t9guge4t0kvqya3q6vamgsejh4q8lxtq9"},{"key":"action","value":"execute_get_next_randomness"}]}]}]';
```
ethanfrey, line 24: Would be nice to include some returns here (easier to read, and parser just parses it as JSON anyway), but fine to leave it also to keep same as actual tx.

ethanfrey (approved):
lgtm

I'd love the 2 packet example, but this is a start.

PR #258, opened 2022-10-19
Fix event parsing

### LINKED ISSUE #255: Sending multiple packets from one tansaction (or msg?) not supported

webmaster128:
Fixes #255

### DIFF
--- a/src/lib/endpoint.ts
+++ b/src/lib/endpoint.ts
@@ -1,4 +1,3 @@
-import { toHex } from '@cosmjs/encoding';
 import { logs } from '@cosmjs/stargate';
 import { tendermint34 } from '@cosmjs/tendermint-rpc';
 import { Packet } from 'cosmjs-types/ibc/core/channel/v1/channel';
@@ -15,8 +14,6 @@ export interface PacketWithMetadata {
   packet: Packet;
   // block it was in, must query proofs >= height
   height: number;
-  // who send the packet (first signer of the tx this was in)
-  sender: string;
 }
 
 export type AckWithMetadata = Ack & {
@@ -99,25 +96,11 @@ export class Endpoint {
     }
 
     const search = await this.client.tm.txSearchAll({ query });
-    const resultsNested = search.txs.map(({ hash, height, result }) => {
+    const resultsNested = search.txs.map(({ height, result }) => {
       const parsedLogs = logs.parseRawLog(result.log);
-      // we accept message.sender (cosmos-sdk) and message.signer (x/wasm)
-      let sender = '';
-      try {
-        sender = logs.findAttribute(parsedLogs, 'message', 'sender').value;
-      } catch {
-        try {
-          sender = logs.findAttribute(parsedLogs, 'message', 'signer').value;
-        } catch {
-          this.client.logger.warn(
-            `No message.sender nor message.signer in tx ${toHex(hash)}`
-          );
-        }
-      }
       return parsePacketsFromLogs(parsedLogs).map((packet) => ({
         packet,
         height,
-        sender,
       }));
     });
     return ([] as PacketWithMetadata[]).concat(...resultsNested);
--- a/src/lib/endpoint.ts
+++ b/src/lib/endpoint.ts
@@ -96,14 +96,16 @@ export class Endpoint {
     }
 
     const search = await this.client.tm.txSearchAll({ query });
-    const resultsNested = search.txs.map(({ height, result }) => {
-      const parsedLogs = logs.parseRawLog(result.log);
-      return parsePacketsFromLogs(parsedLogs).map((packet) => ({
-        packet,
-        height,
-      }));
-    });
-    return ([] as PacketWithMetadata[]).concat(...resultsNested);
+    const resultsNested = search.txs.map(
+      ({ height, result }): PacketWithMetadata[] => {
+        const parsedLogs = logs.parseRawLog(result.log);
+        return parsePacketsFromLogs(parsedLogs).map((packet) => ({
+          packet,
+          height,
+        }));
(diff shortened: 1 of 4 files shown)


src/lib/utils.ts
```diff
@@ -194,22 +193,48 @@ interface ParsedEvent {
   readonly attributes: readonly ParsedAttribute[];
 }
 
+/**
+ * In Tendermint 0.34, event attributes are raw bytes. This changes
```
ethanfrey, line 197: In particular, I believe the ibc_packet data field is binary. Which caused a number of troubles, so they added hex_data field, which was ascii (and I believe what we use). I assume the old binary field is still there.

I would be careful this doesn't throw for protocols outside of ics20 which actually put binary data in the packet data (using the suggestion protobuf, which was NOT used by the ics20 standard for whatever reason - even though they define the format in protobuf 🤷 )

ethanfrey: Links:
* [Deprecated raw packet data field](https://github.com/cosmos/ibc-go/blob/c86d27fc280cfb342a9e4689b381e5823441b694/modules/core/04-channel/keeper/events.go#L134) is still there, using [packet_data](https://github.com/cosmos/ibc-go/blob/c86d27fc280cfb342a9e4689b381e5823441b694/modules/core/04-channel/types/events.go#L26) key

[bytes for outgoing ica packet](https://github.com/cosmos/ibc-go/blob/c86d27fc280cfb342a9e4689b381e5823441b694/modules/apps/27-interchain-accounts/controller/keeper/relay.go#L66) are ... [still proto-JSON](https://github.com/cosmos/ibc-go/blob/c86d27fc280cfb342a9e4689b381e5823441b694/modules/apps/27-interchain-accounts/types/packet.go#L45-L48) even if they are full of complex protobuf types :shrug:

I guess this is safe enough. I only hit the binary protocol with Desmos, who was following ibc team "best practices" which were never used.

ethanfrey (commented):
Generally looks good.

1. can you add a full packet data output test (like the one you just added for raw_log)?
2. we should handle data field being binary without a panic (maybe blank field?). although after investigation major protocols are all using json strings these days

src/lib/endpoint.ts
```diff
@@ -99,28 +96,14 @@ export class Endpoint {
     }
 
     const search = await this.client.tm.txSearchAll({ query });
-    const resultsNested = search.txs.map(({ hash, height, result }) => {
```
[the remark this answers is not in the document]
webmaster128: Actually three reasons. I saw the warning _No message.sender nor message.signer in tx …_ a lot of times in the logs, which confuses people trying to get things up and running. The way the code was structured this was also searched in transactions that do not even have a packet send. Then this provides some signer on the transaction basis, but does not tell much about the actual packages. With multiple signers this can be unrelated to the package emission or the packages are sent after a chain of contract executions. And then the field is not used.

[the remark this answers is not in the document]
ethanfrey: fair enough

webmaster128:
> can you add a full packet data output test (like the one you just added for raw_log)?

Yeah, good point. Can't easily test the RPC output because the Tendermint client already does UTF-8 decoding for me but testing `parsePacketsFromEvents` makes sense.

src/lib/utils.ts
webmaster128: Ah, right. Done in 988da83b7977cc40a4b2f2c1478822752e07cf4a.

webmaster128:
Except for the changelog, this should be done now

src/lib/utils.spec.ts
```diff
@@ -47,6 +49,495 @@ test('stringifyEvent works', (t) => {
...
+        key: fromBase64('YW1vdW50'),
+        value: fromBase64('MzY5NDV1anVub3g='),
+      },
+    ],
+  });
+  t.deepEqual(event2, {
+    type: 'coin_spent',
+    attributes: [
+      {
+        key: '���',
```
ethanfrey, line 73: ah, nice

src/lib/utils.spec.ts
```diff
@@ -47,6 +49,495 @@ test('stringifyEvent works', (t) => {
...
+        },
+      ],
+    },
+  ];
+
+  // See https://testnet.mintscan.io/juno-testnet/txs/F64B8C6A320A9C25FD1EA60B00194817B069C9CBEF19B736117D9339F33F2E51
+  // for packet logs
+  const packets = parsePacketsFromEvents(events);
+  t.is(packets.length, 3);
+  const [packet0, packet1, packet2] = packets;
```
ethanfrey, line 498: awesome! shows getting all packets!

src/lib/utils.ts
```diff
@@ -195,23 +195,31 @@ interface ParsedEvent {
...
   return {
     type,
-    attributes: attributes.map(
-      ({ key, value }): ParsedAttribute => ({
-        key: fromUtf8(key),
-        value: fromUtf8(value),
-      })
-    ),
+    attributes: attributes.map(({ key, value }): ParsedAttribute => {
+      let keyStr = '���';
```
ethanfrey, line 208: Great solution! Loving the unicode chars

ethanfrey (approved):
lgtm. great fix!

PR #263, opened 2022-12-13
Upgrade cosmjs and use lossy fromUtf8 implementation

webmaster128:
This is now implemented in `fromUtf8` directly (and better). No need to have the logic here.


src/lib/utils.ts
```diff
@@ -205,19 +205,10 @@ export function stringifyEvent(event: tendermint34.Event): ParsedEvent {
...
-        // eslint-disable-next-line no-empty
-      } catch {}
-      try {
-        valueStr = fromUtf8(value);
-        // eslint-disable-next-line no-empty
-      } catch {}
       return {
-        key: keyStr,
-        value: valueStr,
+        // Lossy UTF-8 conversion using � replacement characters
```
ethanfrey, line 209: Nice you added it to cosmjs

ethanfrey (approved):
lgtm

PR #264, opened 2023-02-18
Add txHash to AckWithMetadata

webmaster128:
This is useful if you want to learn more about what happened in the ack, e.g. by querying the events of the transaction. We could also add the events directly, but the transaction hash is a bit simpler and more generic.

### DIFF
--- a/src/lib/endpoint.ts
+++ b/src/lib/endpoint.ts
@@ -135,15 +135,15 @@ export class Endpoint {
     }
 
     const search = await this.client.tm.txSearchAll({ query });
-    const resultsNested = search.txs.map(({ height, result }) => {
+    const out = search.txs.flatMap(({ height, result }) => {
       const parsedLogs = logs.parseRawLog(result.log);
       // const sender = logs.findAttribute(parsedLogs, 'message', 'sender').value;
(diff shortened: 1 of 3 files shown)


src/lib/endpoint.ts
```diff
@@ -135,15 +141,18 @@ export class Endpoint {
...
-    const resultsNested = search.txs.map(({ height, result }) => {
+    const out = search.txs.flatMap(({ height, result, hash }) => {
       const parsedLogs = logs.parseRawLog(result.log);
       // const sender = logs.findAttribute(parsedLogs, 'message', 'sender').value;
-      return parseAcksFromLogs(parsedLogs).map((ack) => ({
-        height,
-        ...ack,
-      }));
+      return parseAcksFromLogs(parsedLogs).map(
+        (ack): AckWithMetadata => ({
```
ethanfrey, line 148: Better place to annotate

src/lib/endpoint.ts
```diff
@@ -135,15 +141,18 @@ export class Endpoint {
...
+          height,
+          txHash: toHex(hash).toUpperCase(),
```
ethanfrey, line 150: nice addition

ethanfrey (approved):
lgtm

PR #265, opened 2023-03-07
Upgrade CosmJS to 0.30

webmaster128:


webmaster128:
> lgtm? why stuck?

No particular reason. Just not yet reviewed. Do you need this PR or a ts-relayer release with ComsJS 0.30?

PR #270, opened 2023-06-22
Upgrade CosmJS to 0.31 and cosmjs-types to 0.8.0

webmaster128:


src/lib/ibcclient.ts
```diff
@@ -705,7 +705,7 @@ export class IbcClient {
       value: MsgUpdateClient.fromPartial({
         signer: senderAddress,
         clientId,
-        header: {
+        clientMessage: {
```
ethanfrey, line 708: This only works with newer versions of the SDK then?
It would be good to document compatibility

ethanfrey (approved):
Looks good. Wonder about compatibility with the protobuf message

src/lib/ibcclient.ts
webmaster128: The field was renamed in https://github.com/cosmos/ibc-go/pull/1871. But since it is the same field number in the protobuf serialization as before I assume nothing bad happens with the change 🤞

ethanfrey: Only difference is amino signing, then? Which likely isn't relevant here

PR #275, opened 2023-11-28
feat: Packet-level filtering

clockworkgr:
This PR adds the option to only relay specific packets that are of interest to the app developer.

In order to avoid unnecessary development and maintenance overhead, the filtering mechanism is left entirely to the app developer.

This is achieved by defining a new function type: `PacketFilter` that takes a `Packet` as an argument and returns `true` if it should be relayed and `false` if it should be ignored.

Implementation of this function is left to the application developer and can be used by setting it via `link.setFilter(packetFilter: PacketFilter)`  on the `Link` instance. 

This PR also includes the complementary `Link.clearFilter()` method.

### DIFF
--- a/src/lib/link.ts
+++ b/src/lib/link.ts
@@ -38,7 +38,7 @@ export function otherSide(side: Side): Side {
   }
 }
 /**
- * PacketFilter is the type for a function that accepts a Packet and returns a boolean definign whether to relay the packet or not
+ * PacketFilter is the type for a function that accepts a Packet and returns a boolean defining whether to relay the packet or not
  */
 
 export type PacketFilter = (packet: Packet) => boolean;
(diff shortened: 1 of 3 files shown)


src/lib/link.ts
```diff
@@ -37,6 +37,11 @@ export function otherSide(side: Side): Side {
     return 'A';
   }
 }
+/**
+ * PacketFilter is the type for a function that accepts a Packet and returns a boolean definign whether to relay the packet or not
+ */
+
+export type PacketFilter = (packet: Packet) => boolean;
```
webmaster128, line 44: This empty line between doc comment and type seems off. It probably will cause the doc comment to not be found

webmaster128 (commented):
Filtering looks good to me, just the newline

PR #277, opened 2023-11-29
fix: Ensure CommitSig Timestamp is empty if BlockIdFlag is ABSENT -- DO_NOT_MERGE

clockworkgr:


clockworkgr:
Best if we close this in favour of https://github.com/cosmos/cosmjs/pull/1516 (assuming it works, which it should)

clockworkgr:
> Best if we close this in favour of [cosmos/cosmjs#1516](https://github.com/cosmos/cosmjs/pull/1516) (assuming it works, which it should)

Tested [cosmos/cosmjs#1516](https://github.com/cosmos/cosmjs/pull/1516) and it works. Probably a better solution

clockworkgr:
Closing in favour of https://github.com/confio/ts-relayer/pull/278

PR #279, opened 2023-12-06
feat: Switch to events parsing

clockworkgr:
Raw logs have been deprecated in SDK 0.50.0 so switching to `result.events` to get IBC client/connection/channel data.


webmaster128:
Much nicer that way :)

PR #285, opened 2024-03-07
Fix header validation error

NoahSaso:
I think I solved the infamous `header failed basic validation` error!

```
Query failed with (6): rpc error: code = Unknown desc = header failed basic validation: commit signs block 333F5ECBE51CA88088C64523E7CF156A67695D949ABC23E02270FA57EF2D4269, header is block 444DB206D17B81B86392FCB24AF2162B956EBC5D7FBE367ACE3213545AAD926B [cosmos/ibc-go/v6@v6.2.0/modules/light-clients/07-tendermint/types/header.go:62] With gas wanted: '0' and gas used: '46662' : unknown request
```

It appears a field of the header was not being copied over correctly. Presumably it would succeed when `rpcHeader.version.app` was 0, and fail the rest of the time.

### DIFF
--- a/.changeset/thick-clouds-admire.md
+++ b/.changeset/thick-clouds-admire.md
@@ -2,4 +2,4 @@
 "@confio/relayer": patch
 ---
 
-fix header failed basic validation error
+fix header failed basic validation error for app version != 0
--- a/.changeset/thick-clouds-admire.md
+++ b/.changeset/thick-clouds-admire.md
@@ -1,0 +1,5 @@
+---
+"@confio/relayer": patch
+---
+
+fix header failed basic validation error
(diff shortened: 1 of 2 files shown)


webmaster128 (approved):
Looks correct to me. If this helps, great!

NoahSaso:
> Looks correct to me. If this helps, great!

sweet! let me know when we can merge and publish a new version :)

seems like the CI isn't running

.changeset/thick-clouds-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@confio/relayer": patch
+---
+
+fix header failed basic validation error
```
webmaster128, line 5:
```suggestion
fix header failed basic validation error for app version != 0
```

webmaster128:
Can you try version 0.12.0?

noahsaso:
> Can you try version 0.12.0?

yep, one minute

noahsaso:
> > Can you try version 0.12.0?
> 
> yep, one minute

first two tests succeeded!

it's hard to verify because it was never deterministic, but the fact that two succeeded in a row was very rare before, so it seems to be working 🥳

